PR #2, opened 2024-12-13
build: Add dependabot config for automatic dependency updates.

klkucaj:


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "pub"
+    directory: "/"
```
SandPod, line 4: Question: Is this required?

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
...
+    schedule:
+      interval: "daily"
+    versioning-strategy: "widen"
+    # Limit open pull requests to 1 to avoid multiple PRs for the same dependency.
+    open-pull-requests-limit: 1
```
SandPod, line 9: Suggestion: Set this to 5, it is OK to have multiple requests open here.

Update the comment as well.

.github/dependabot.yml
klkucaj: yes it is required https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

Screenshot 2024-12-13 at 12 15 05 PM

SandPod (approved):
LGTM! 👍

PR #17, opened 2024-12-17
feat: Implement Lazy Loading for Header Parsing to Avoid Unnecessary Validation

### LINKED ISSUE #10: RELIC: Implement Lazy Loading for Header Parsing to Avoid Unnecessary Validation
**Is your feature request related to a problem? Please describe.**  
Currently, headers such as `Accept-Language` are parsed and validated eagerly, even if the request handler does not actually read or use them. This can lead to unnecessary rejections of requests with invalid headers, even though those headers are irrelevant to the processing logic. This behavior might not align with how other server frameworks handle unused headers, as they may only parse or validate headers on demand.

**Describe the solution you'd like**  
Implement lazy loading for header parsing and validation. Instead of parsing all headers upfront, headers should only be parsed or validated when explicitly accessed by the request handler. This approach can:
- Improve performance by skipping unnecessary work.
- Avoid rejecting requests for irrelevant headers.
- Provide behavior more in line with other server frameworks.

**Describe alternatives you've considered**  
1. **Eager Parsing with Error Suppression**: Continue parsing headers eagerly but suppress validation errors for unused headers. However, this might lead to silent failures and make debugging harder.
2. **Selective Parsing via Configuration**: Allow developers to configure which headers to validate. This adds complexity and requires developers to be proactive about configurations.

**Additional context**  
The current behavior raises questions about the wisdom of rejecting requests based on irrelevant headers. This issue aligns with making the framework more efficient and user-friendly. Lazy loading is a common pattern used in many frameworks to avoid unnecessary overhead.

Feedback: https://github.com/serverpod/serverpod/pull/2744#discussion_r1873254470

klkucaj:
### **PR Description**  
This PR implements lazy loading for header parsing and validation. Headers are now parsed and validated only when explicitly accessed by the request handler, improving performance and avoiding unnecessary validation for unused headers.  

Closes #10

### DIFF
--- a/lib/src/extensions/http_response_extension.dart
+++ b/lib/src/extensions/http_response_extension.dart
@@ -14,7 +14,11 @@ extension HttpResponseExtension on io.HttpResponse {
     // Apply all headers from the provided headers map.
     var mappedHeaders = headers.toMap();
     for (var entry in mappedHeaders.entries) {
-      responseHeaders.set(entry.key, entry.value);
+      var key = entry.key;
+      var value = entry.value;
+      if (value != null) {
+        responseHeaders.set(key, value);
+      }
     }
 
     // Set Content-Type based on the MIME type of the body.
--- a/lib/src/headers/headers.dart
+++ b/lib/src/headers/headers.dart
@@ -1364,12 +1364,9 @@ abstract base class Headers {
   });
 
   /// Convert headers to a map
-  ///
-  /// If [allowNullValueToBeIncluded] is `true`, then null values will be
-  /// included in the map, otherwise they will be excluded.
-  Map<String, Object?> toMap({
-    bool allowNullValueToBeIncluded = false,
-  });
+  /// This will include all headers, if a header is null then the value of the
+  /// header was not set.
+  Map<String, Object?> toMap();
 }
 
 /// Headers implementation
@@ -1719,14 +1716,8 @@ final class _HeadersImpl extends Headers {
     );
   }
 
-  /// Convert headers to a map
-  ///
-  /// If [allowNullValueToBeIncluded] is `true`, then null values will be
-  /// included in the map, otherwise they will be excluded.
   @override
-  Map<String, Object?> toMap({
-    bool allowNullValueToBeIncluded = false,
-  }) {
+  Map<String, Object?> toMap() {
     var map = <String, Object?>{};
 
     // Date-related headers
@@ -1734,69 +1725,43 @@ final class _HeadersImpl extends Headers {
     for (var entry in dateHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null || allowNullValueToBeIncluded) {
-        map[key] = value != null ? formatHttpDate(value) : null;
-      }
+      map[key] = value != null ? formatHttpDate(value) : null;
     }
 
     // Number-related headers
     var numberHeaders = _numberHeadersMap;
     for (var entry in numberHeaders.entries) {
-      var key = entry.key;
-      var value = entry.value;
-      if (value != null || allowNullValueToBeIncluded) {
-        map[key] = value;
-      }
+      map[entry.key] = entry.value;
     }
 
     // Boolean-related headers
     var booleanHeaders = _booleanHeadersMap;
     for (var entry in booleanHeaders.entries) {
-      var key = entry.key;
-      var value = entry.value;
-      if (value != null || allowNullValueToBeIncluded) {
-        map[key] = value;
-      }
+      map[entry.key] = entry.value;
     }
 
     // String-related headers
     var stringHeaders = _stringHeadersMap;
     for (var entry in stringHeaders.entries) {
-      var key = entry.key;
-      var value = entry.value;
-      if (value != null || allowNullValueToBeIncluded) {
-        map[key] = value;
-      }
+      map[entry.key] = entry.value;
     }
 
     // List<String>-related headers
     var listStringHeaders = _listStringHeadersMap;
     for (var entry in listStringHeaders.entries) {
-      var key = entry.key;
-      var value = entry.value;
-      if (value != null || allowNullValueToBeIncluded) {
-        map[key] = value;
-      }
+      map[entry.key] = entry.value;
     }
 
     // Uri-related headers
     var uriHeaders = _uriHeadersMap;
     for (var entry in uriHeaders.entries) {
-      var key = entry.key;
-      var value = entry.value;
-      if (value != null || allowNullValueToBeIncluded) {
-        map[key] = value.toString();
-      }
+      map[entry.key] = entry.value?.toString();
     }
 
     // TypedHeader-related headers
     var typedHeaders = _typedHeadersMap;
     for (var entry in typedHeaders.entries) {
-      var key = entry.key;
-      var value = entry.value;
-      if (value != null || allowNullValueToBeIncluded) {
-        map[key] = value?.toHeaderString();
-      }
+      map[entry.key] = entry.value?.toHeaderString();
     }
 
     // Custom headers
--- a/lib/src/headers/headers.dart
+++ b/lib/src/headers/headers.dart
@@ -315,7 +315,7 @@ abstract base class Headers {
 
   /// Managed headers
   /// Headers that are managed by the library
-  static const _managedHeaders = <String>{
+  static const managedHeaders = <String>{
     dateHeader,
     expiresHeader,
     ifModifiedSinceHeader,
@@ -999,7 +999,7 @@ abstract base class Headers {
       // Custom Headers
       custom: parseCustomHeaders(
         dartIOHeaders.headers,
-        excludedHeaders: _managedHeaders,
+        excludedHeaders: managedHeaders,
       ),
 
       // Failed Headers to Parse
@@ -1201,7 +1201,6 @@ abstract base class Headers {
       lastModified: _LazyInit.withValue(lastModified),
       ifModifiedSince: _LazyInit.nullValue(),
       ifUnmodifiedSince: _LazyInit.nullValue(),
-
       // General Headers
       origin: _LazyInit.withValue(origin),
       server: _LazyInit.withValue(server),
@@ -1365,7 +1364,12 @@ abstract base class Headers {
   });
 
   /// Convert headers to a map
-  Map<String, Object> toMap();
+  ///
+  /// If [allowNullValueToBeIncluded] is `true`, then null values will be
+  /// included in the map, otherwise they will be excluded.
+  Map<String, Object?> toMap({
+    bool allowNullValueToBeIncluded = false,
+  });
 }
 
 /// Headers implementation
@@ -1716,17 +1720,22 @@ final class _HeadersImpl extends Headers {
   }
 
   /// Convert headers to a map
+  ///
+  /// If [allowNullValueToBeIncluded] is `true`, then null values will be
+  /// included in the map, otherwise they will be excluded.
   @override
-  Map<String, Object> toMap() {
-    var map = <String, Object>{};
+  Map<String, Object?> toMap({
+    bool allowNullValueToBeIncluded = false,
+  }) {
+    var map = <String, Object?>{};
 
     // Date-related headers
     var dateHeaders = _dateHeadersMap;
     for (var entry in dateHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null) {
-        map[key] = formatHttpDate(value);
+      if (value != null || allowNullValueToBeIncluded) {
+        map[key] = value != null ? formatHttpDate(value) : null;
       }
     }
 
@@ -1735,7 +1744,7 @@ final class _HeadersImpl extends Headers {
     for (var entry in numberHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null) {
+      if (value != null || allowNullValueToBeIncluded) {
         map[key] = value;
       }
     }
@@ -1745,7 +1754,7 @@ final class _HeadersImpl extends Headers {
     for (var entry in booleanHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null) {
+      if (value != null || allowNullValueToBeIncluded) {
         map[key] = value;
       }
     }
@@ -1755,7 +1764,7 @@ final class _HeadersImpl extends Headers {
     for (var entry in stringHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null) {
+      if (value != null || allowNullValueToBeIncluded) {
         map[key] = value;
       }
     }
@@ -1765,7 +1774,7 @@ final class _HeadersImpl extends Headers {
     for (var entry in listStringHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null) {
+      if (value != null || allowNullValueToBeIncluded) {
         map[key] = value;
       }
     }
@@ -1775,7 +1784,7 @@ final class _HeadersImpl extends Headers {
     for (var entry in uriHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null) {
+      if (value != null || allowNullValueToBeIncluded) {
         map[key] = value.toString();
       }
     }
@@ -1785,8 +1794,8 @@ final class _HeadersImpl extends Headers {
     for (var entry in typedHeaders.entries) {
       var key = entry.key;
       var value = entry.value;
-      if (value != null) {
-        map[key] = value.toHeaderString();
+      if (value != null || allowNullValueToBeIncluded) {
+        map[key] = value?.toHeaderString();
       }
     }
 
--- a/lib/src/headers/headers.dart
+++ b/lib/src/headers/headers.dart
@@ -116,29 +116,74 @@ abstract base class Headers {
   List<String>? get via => _via.value;
 
   /// Request Headers
-  final FromHeader? from;
-  final Uri? host;
-  final AcceptEncodingHeader? acceptEncoding;
-  final AcceptLanguageHeader? acceptLanguage;
-  final List<String>? accessControlRequestHeaders;
-  final RequestMethod? accessControlRequestMethod;
-  final int? age;
-  final List<RequestMethod>? allow;
-  final AuthorizationHeader? authorization;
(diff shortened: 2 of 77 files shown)


lib/src/relic_server.dart
```diff
@@ -175,6 +183,10 @@ class RelicServer {
           ),
         );
       }
+    } on InvalidHeaderException catch (error) {
```
SandPod, line 186: Question: This exception is exported right?

I'm inspecting the first commit; we could probably have added tests that validate that the changes in that commit work if the exception was exported.

lib/src/relic_server.dart
```diff
@@ -175,6 +183,10 @@ class RelicServer {
...
+      return Response.badRequest(
```
SandPod, line 187: Question: Should't we create a log entry here as well? We could end up here because someone copied the response headers right?

lib/src/headers/headers.dart
```diff
@@ -260,83 +394,171 @@ abstract base class Headers {
...
-    this.accessControlAllowMethods,
-    this.accessControlAllowHeaders,
-    this.clearSiteData,
-    this.secFetchDest,
-    this.secFetchMode,
-    this.secFetchSite,
-    this.crossOriginResourcePolicy,
-    this.crossOriginEmbedderPolicy,
-    this.crossOriginOpenerPolicy,
+    // Failed headers to parse
```
SandPod, line 481: Suggestion: Should we rename this to `invalidHeaders`?  This should be done in another PR if we should to it.

lib/src/headers/headers.dart
```diff
@@ -1386,7 +1884,57 @@ final class _HeadersImpl extends Headers {
...
+  }) {
+    return _LazyInit._(
+      value: value,
+      isInitialized: true,
+    );
+  }
+
+  factory _LazyInit.lazy({
+    required _LazyInitializer<T> init,
+  }) {
```
SandPod, line 1920: Suggestion: Rename to `_LazyInit.from` and replace the named named parameter with a mandatory parameter, this factory can simply take a single argument.

lib/src/headers/headers.dart
```diff
@@ -1386,7 +1884,57 @@ final class _HeadersImpl extends Headers {
...
+
+  _LazyInit._({
+    _LazyInitializer<T>? init,
+    T? value,
+    bool isInitialized = false,
+  })  : _init = init,
+        _value = value,
+        _isInitialized = isInitialized;
+
+  factory _LazyInit.value({
```
SandPod, line 1909: Suggestion: rename to `_LazyInit.withValue` and replace the named parameter with a single mandatory argument.

test/headers/headers_test_utils.dart
```diff
@@ -39,12 +39,21 @@ Future<RelicServer> createServer({
...
 }) async {
   Headers? parsedHeaders;
 
   server.mountAndStart(
     (Request request) {
       parsedHeaders = request.headers;
+
+      if (parseAllHeaders) {
+        // This will force the headers to be parsed and the lazy parsing to
+        // complete.
```
SandPod, line 53: Suggestion: Remove comment, this is self-explanatory from the variable name.

lib/src/headers/headers.dart
```diff
@@ -1735,6 +1735,16 @@ final class _HeadersImpl extends Headers {
       }
     }
 
+    // Boolean-related headers
+    var booleanHeaders = _booleanHeadersMap;
+    for (var entry in booleanHeaders.entries) {
```
SandPod, line 1740: Question: Was this missing before? Should we have a test that validates that we actually add all headers?

test/headers/typed_headers/transfer_encoding_header_test.dart
```diff
@@ -54,31 +54,31 @@ void main() {
       },
     );
 
-    /// According to the HTTP/1.1 specification (RFC 9112), the 'chunked' transfer
-    /// encoding must be the final encoding applied to the response body.
     test(
-      'when a valid Transfer-Encoding header is passed with "chunked" as not the last '
-      'encoding then it should parse the encodings correctly and reorder them sot the '
-      'chunked encoding is the last encoding',
+      'when a Transfer-Encoding header with an invalid value is passed '
```
SandPod, line 58: Question: Why was a test removed in this commit? https://github.com/serverpod/relic/pull/17/commits/c95c70c67e89d0e184945a0407f586172299e9f6

lib/src/headers/headers.dart
klkucaj: We don’t need that because we already have tests covering all headers. I added this specifically because the tests for this header were failing. Previously, these tests were failing along with many others due to the transfer encoding and content length issue. Now that the issue has been resolved and merged, I had a clearer understanding of what needed to be fixed and what was already addressed in the other ticket.

lib/src/relic_server.dart
klkucaj: yes, done

lib/src/headers/headers.dart
klkucaj: Now this depends, `invalidHeaders` is a good naming but also `failedHeadersToParse` tells we failed to parse these headers, but still i am not sure, what do you suggest ?

lib/src/relic_server.dart
```diff
@@ -127,42 +127,50 @@ class RelicServer {
...
-      } else {
-        errorResponse = Response.internalServerError();
+      if (error is ArgumentError) {
+        if (error.name == 'method' || error.name == 'requestedUri') {
+          return Response.badRequest().writeHttpResponse(request.response);
+        }
       }
 
       // Write the response to the HTTP response.
-      await errorResponse.writeHttpResponse(
```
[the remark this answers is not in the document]
klkucaj: Nope it was not dead code, check this:
Before:
https://github.com/serverpod/relic/blob/e2607bc2244bea41867f64c2971714416df9b407/lib/src/relic_server.dart#L143-L163

After:
https://github.com/klkucaj/relic/blob/fdd40a728760f4fe5559d8a000038c0f75e0cc3d/lib/src/relic_server.dart#L141-L170


Previously, we created the response first and then wrote it. Now, we handle each condition individually as we write the response.

test/headers/typed_headers/transfer_encoding_header_test.dart
klkucaj: I remove this:
https://github.com/serverpod/relic/blob/e2607bc2244bea41867f64c2971714416df9b407/test/headers/typed_headers/transfer_encoding_header_test.dart#L76-L89

because its basically same as this:
https://github.com/serverpod/relic/blob/e2607bc2244bea41867f64c2971714416df9b407/test/headers/typed_headers/transfer_encoding_header_test.dart#L57-L74

Should i add it back ?

lib/src/relic_server.dart
klkucaj: Check this commit and let me know: https://github.com/serverpod/relic/pull/17/commits/de47c92c2626827d348d508d76b6f6367528c0c3

lib/src/headers/headers.dart
SandPod: This has not been fixed.

lib/src/headers/headers.dart
SandPod: If a header is added but forgotten to be added to any of the maps. Would some form of test catch that? Like a response request test or something?

lib/src/headers/headers.dart
klkucaj: Done

test/headers/typed_headers/transfer_encoding_header_test.dart
SandPod: To me, these tests seems to serve different purposes. They have different configuration. The one bellow validates that the entries are sorted correctly. The top one does not take that into consideration for its validation. Is there any other test that you consider overlaps with the top one? Otherwise, I don't think this motivates removing the test.

test/exception/relic_exceptions_test.dart
```diff
@@ -0,0 +1,112 @@
...
+      );
+      final response = await _get();
+      expect(response.statusCode, 501);
+      expect(response.body, 'Not Implemented');
+    });
+
+    test(
+        'when a handler throws an HijackException '
+        'then it returns a 500 Internal Server Error response', () async {
+      await _scheduleServer((Request request) => throw HijackException());
```
SandPod, line 56: Question: Isn't the purpose of the Hijack exception to take over what happens with the request?

test/exception/relic_exceptions_test.dart
```diff
@@ -0,0 +1,112 @@
...
+      } finally {
+        _server = null;
+      }
+    }
+  });
+
+  group('Given a server', () {
+    test(
+        'when a handler throws an InvalidHeaderException '
+        'then it returns a 400 Bad Request response', () async {
```
SandPod, line 30: Suggestion: Change to `then it returns a 400 Bad Request response with exception message included in body.`

SandPod (changes requested):

lib/src/headers/headers.dart
klkucaj: If a new header is added without including the corresponding test for the header, it will not be detected. Unfortunately, there doesn't seem to be a way to avoid this entirely. The key idea is that whenever a header is added, it must also be included in the map.

test/headers/typed_headers/transfer_encoding_header_test.dart
klkucaj: I think its better to add it back.

test/exception/relic_exceptions_test.dart
klkucaj: Yes, that's correct. However, when you take over (hijack) a request, you should not attempt to write it back as a response. Instead, you need to manage the socket stream directly to write, read, or close the connection as needed.

lib/src/headers/headers.dart
SandPod: Could you file an issue for this? I think this can be solved by simply providing a way to retrive all the headers from `toMap` without filtering out the null values.
But we would also need to have a way to automatically iterate all of the header constants.

lib/src/relic_server.dart
```diff
@@ -127,42 +127,47 @@ class RelicServer {
...
+    } on InvalidHeaderException catch (error, stackTrace) {
+      // If the request headers are invalid, respond with a 400 Bad Request status.
+      logMessage(
+        'Error parsing request headers.\n$error',
+        stackTrace: stackTrace,
+        type: LoggerType.error,
+      );
+      // Write the response to the HTTP response.
+      return Response.badRequest(
+        body: Body.fromString(error.toString()),
```
SandPod, line 147: Suggestion: Add a different field to the `InvalidHeaderException` that is used to build the body of the request.

The purpose of this is to clearly differentiate between internal bug messages compared to what we send to the client.

lib/src/relic_server.dart
```diff
@@ -185,24 +200,47 @@ class RelicServer {
...
-      response = Response.internalServerError();
+      return Response.internalServerError().writeHttpResponse(
+        request.response,
+      );
     }
 
     // If the the request is already hijacked, meaning it's being handled by
     // another handler, like a websocket, then respond with a 501 Not Implemented status.
     // This should never happen, but if it does, we don't want to respond with an error.
     if (!relicRequest.canHijack) {
```
SandPod, line 220: Question: When do we end up in this case? Don't we already handle the hijack above? Feels like this will hit as soon as you disable hijack in for relic requests?

test/exception/relic_exceptions_test.dart
SandPod: But this only happens if hijack is disabled right? If it is enabled we shouldn't throw if I understand the code correctly? So that should be reflected in the given and part of the setup for the test.

SandPod (changes requested):

lib/src/relic_server.dart
klkucaj: Fixed!

lib/src/headers/headers.dart
```diff
@@ -102,6 +102,9 @@ abstract base class Headers {
   final _LazyInit<DateTime?> _ifModifiedSince;
   DateTime? get ifModifiedSince => _ifModifiedSince.value;
 
+  final _LazyInit<DateTime?> _ifUnmodifiedSince;
```
SandPod, line 105: Comment: This is a feature that should have been added in a different PR. It has nothing to do with lazy loading.

But keep it now.

lib/src/extensions/http_response_extension.dart
```diff
@@ -14,7 +14,11 @@ extension HttpResponseExtension on io.HttpResponse {
     // Apply all headers from the provided headers map.
     var mappedHeaders = headers.toMap();
     for (var entry in mappedHeaders.entries) {
-      responseHeaders.set(entry.key, entry.value);
+      var key = entry.key;
+      var value = entry.value;
+      if (value != null) {
```
SandPod, line 19: Question/Discussion: Now we will always filter out `null` custom headers as well, restricting the flexibility of the users of this interface. Do we really want to do that?

lib/src/headers/headers.dart
```diff
@@ -1365,7 +1364,12 @@ abstract base class Headers {
   });
 
   /// Convert headers to a map
-  Map<String, Object> toMap();
+  ///
+  /// If [allowNullValueToBeIncluded] is `true`, then null values will be
+  /// included in the map, otherwise they will be excluded.
+  Map<String, Object?> toMap({
+    bool allowNullValueToBeIncluded = false,
```
SandPod, line 1371: Suggestion: This only applies to managed headers. Please update the message and variable name accordingly.

test/headers/header_test.dart
```diff
@@ -110,5 +110,69 @@ void main() {
...
       expect(headers.date, equals(newDate));
     });
+
+    test(
+        'when calling "toMap" without setting the "allowNullValueToBeIncluded" flag'
+        'then "allowNullValueToBeIncluded" will default to false and null values are excluded',
+        () {
+      var headers = Headers.request(age: null);
+      var map = headers.toMap();
+      expect(map.containsKey('age'), isFalse);
```
SandPod, line 120: Comment: This needs an explanation, why does this key not being present imply that null values are excluded? Update the description.

test/headers/header_test.dart
```diff
@@ -110,5 +110,69 @@ void main() {
       headers = headers.copyWith(date: newDate);
       expect(headers.date, equals(newDate));
     });
+
+    test(
```
SandPod, line 114: Suggestion: For these tests, part of the given is that we have a headers class that does not have any headers added to it. I suggest you break this out into a separate group so that this is clear.

lib/src/relic_server.dart
```diff
@@ -214,12 +214,17 @@ class RelicServer {
...
+    //
+    // If the request is already hijacked (e.g., taken over by another handler
+    // like a WebSocket), this is an unexpected state. Typically, hijacking
+    // should throw a [HijackException], but this check serves as a safeguard
+    // for improper hijacking scenarios.
+    if (relicRequest.isHijacked) {
+      throw StateError(
+        'Cannot proceed: the request has already been hijacked by another handler '
+        '(e.g., a WebSocket). Ensure you dont attempt to write to the HTTP response '
+        'after the request has been hijacked.',
```
SandPod, line 227: Suggestion: Remove the last part since that doesn't make us end up here right? `Ensure you dont attempt to write to the HTTP after the request has been hijacked.`.

And give a proper error message to why we ended up here, same as described above. We are missing an expected throw with a `HijackException`.

test/hijack/relic_hijack_test.dart
```diff
@@ -0,0 +1,115 @@
+@TestOn('vm')
+library;
```
SandPod, line 2: Question: Are these required? We don't test on any other platform right?

test/hijack/relic_hijack_test.dart
```diff
@@ -0,0 +1,115 @@
...
+      expect(
+        () async => await _get(),
+        throwsA(isA<TimeoutException>()),
+      );
+    });
+
+    test(
+        'when request is hijacked but no HijackException is thrown '
+        'then server throws a StateError when trying to write the response '
+        'to the HTTP response', () async {
```
SandPod, line 51: Suggestion: Shouldn't we respond with a 500 here as well?

lib/src/extensions/http_response_extension.dart
klkucaj: This is needed because set method does not accept null values.

test/hijack/relic_hijack_test.dart
klkucaj: Removed all

test/headers/header_test.dart
```diff
@@ -111,68 +111,20 @@ void main() {
...
-      var map = headers.toMap(allowNullValueToBeIncluded: true);
-      expect(
-        map.keys.every((header) => Headers.managedHeaders.contains(header)),
-        isFalse,
-      );
+      var managedHeaders = Headers.managedHeaders
+          .where(
+            (header) =>
+                // These headers are not managed by the Headers class but are
+                // managed by the body class and applied later to the response.
```
SandPod, line 120: Suggestion: Include this in the description of the test.

SandPod: when converting to map then all configurable managed headers are included.

test/headers/header_test.dart
```diff
@@ -110,5 +110,51 @@ void main() {
...
+        var missingManagedHeaders = managedHeaders.difference(mapKeys);
+
+        expect(
+          missingManagedHeaders.isEmpty,
+          isTrue,
+          reason: 'Missing managed headers: $missingManagedHeaders.',
+        );
+      });
+
+      test('then no unexpected additional headers are included', () {
```
SandPod, line 137:
```suggestion
      test('then only managed headers are included.', () {
```

test/headers/header_test.dart
```diff
@@ -110,5 +110,51 @@ void main() {
...
+        var headers = Headers.request();
+        var managedHeaders = Headers.managedHeaders
+            .where(
+              (header) =>
+                  // These headers are not managed by the Headers class but are
+                  // managed by the body class and applied later to the response.
+                  header != Headers.contentLengthHeader &&
+                  header != Headers.contentTypeHeader,
```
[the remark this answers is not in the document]
klkucaj: yes

SandPod (approved):

Issue #20, opened 2024-12-17
fix: Content-Length Header Conflict When Using Transfer-Encoding: chunked


klkucaj:
The HTTP specification (RFC 7230) states that when `Transfer-Encoding: chunked` is used, the `Content-Length` header **must not be present** in the same message. Including both headers causes ambiguity in how the body length is determined and violates the HTTP/1.1 standard.

**Current Behavior:**  
When the `Transfer-Encoding` header includes `chunked` (alone or in combination with other encodings like `gzip`), the `Content-Length` header is still being sent in some cases. This results in undefined behavior and may cause protocol errors for HTTP clients or servers.

**Expected Behavior:**  
- If `Transfer-Encoding: chunked` is present, the `Content-Length` header should be omitted.  
- For other `Transfer-Encoding` values (e.g., `gzip`, `compress`, `deflate`) without `chunked`, the `Content-Length` can be included.

**Examples of Valid Cases:**
✅ `Transfer-Encoding: chunked` (no `Content-Length`)  
✅ `Transfer-Encoding: gzip` + `Content-Length: 100`  
✅ `Transfer-Encoding: deflate` + `Content-Length: 256`  

**Examples of Invalid Cases:**
❌ `Transfer-Encoding: chunked` + `Content-Length: 100`  
❌ `Transfer-Encoding: gzip, chunked` + `Content-Length: 150`  

### **Steps to Reproduce:**
1. Send an HTTP request or response with `Transfer-Encoding: chunked` **and** `Content-Length` set.  
2. Observe that both headers are sent.  

**Impact:**  
This issue violates HTTP/1.1 specifications and can lead to inconsistent handling of the response body by clients or servers, causing errors or unexpected behaviors.

### **Proposed Fix:**  
- Add a validation step to ensure that if `Transfer-Encoding` contains `chunked`, the `Content-Length` header is **not included**.  
- Log or throw an error if both headers are detected.

nielsenko:
To solve this issue require relic to know about the dependency between `TransferEncoding` and `ContentLength` headers. There may well be other such cases of invalid combinations, and if we take the responsibility for one we should do for all.

Also, as users can add there own (typed) headers, we should allow users to setup such validation rules as well.

Next, these rules would need to be enforced runtime, which adds runtime overhead to guard against a coding error. Obviously we could use asserts to avoid the cost of validation in production.

[...]

PR #22, opened 2024-12-18
fix: Resolve Content-Length header conflict with Transfer-Encoding: chunked

klkucaj:
This PR resolves an issue where `Content-Length` was incorrectly included alongside `Transfer-Encoding: chunked`.

**Fixes Implemented:**
- Removed `Content-Length` when `Transfer-Encoding: chunked` is applied.
- Ensured `Content-Length` is only included for other valid `Transfer-Encoding` values (e.g., gzip, deflate).
- Added logic to reorder transfer encodings, ensuring `chunked` is always the last encoding.

**Impact:**
This fix ensures compliance with HTTP/1.1, preventing ambiguity and avoiding potential protocol errors in HTTP clients and servers.

**Closes:** #20


test/relic_server_serve_test.dart
```diff
@@ -499,7 +499,7 @@ void main() {
         response.headers,
         containsPair(HttpHeaders.transferEncodingHeader, 'chunked'),
       );
-      expect(response.body, equals('hi'));
+      expect(response.body, equals('2\r\nhi\r\n0\r\n\r\n'));
```
SandPod, line 502: Suggestion: Could you clarify the description on this test so that it is clear what and why we do this validation?

test/message/body_test.dart
```diff
@@ -0,0 +1,186 @@
...
+          headers: Headers.response(
+            transferEncoding: TransferEncodingHeader(
+              encodings: [TransferEncoding.chunked],
+            ),
+          ),
+        ),
+      );
+
+      var response = await _get();
+      expect(response.headers['transfer-encoding'], contains('chunked'));
```
SandPod, line 128: Suggestion: Use `TransferEncoding.chunked` in the validation for the tests.

SandPod: Use this for all tests when referring to any header string.

SandPod (changes requested):
I know a lot of digging went into this one so it will be nice to get this resolved.

I had a couple of comments and suggestion. Maybe we could have a call if anything is unclear.

test/headers/headers_test_utils.dart
```diff
@@ -17,13 +17,16 @@ class BadRequestException implements Exception {
...
+  bool echoHeaders = true,
 }) async {
   Headers? parsedHeaders;
 
   server.mountAndStart(
     (Request request) {
       parsedHeaders = request.headers;
-      return Response.ok();
+      return Response.ok(
+        headers: echoHeaders ? parsedHeaders : null,
```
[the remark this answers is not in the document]
klkucaj: No, there's no difference. Whether you explicitly set `headers` to `null` or just call `Response.ok()` without modifying it, the outcome is the same. The constructor handles it like this:  

```dart
...
headers: headers ?? Headers.response(),
...
```

This means if `headers` is `null`, it will default to `Headers.response()`.

lib/src/headers/typed/headers/transfer_encoding_header.dart
```diff
@@ -42,6 +43,35 @@ class TransferEncodingHeader implements TypedHeader {
...
+  ///
+  /// This function reorders the encodings to comply with the standard and
+  /// ensures compatibility with HTTP clients and intermediaries.
+  static List<TransferEncoding> _reorderEncodings(
+    List<TransferEncoding> encodings,
+  ) {
+    final TransferEncoding? chunked = encodings.firstWhereOrNull(
+      (e) => e.name == TransferEncoding.chunked.name,
+    );
+    if (chunked == null) return encodings;
```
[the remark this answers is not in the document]
klkucaj: The reason I used `firstWhereOrNull` instead of `.contains` is that `.contains` only works directly with primitive types or objects that have properly overridden the `==` operator and `hashCode`. 
In this case, since `TransferEncoding` is a custom class and doesn’t inherently support those overrides, `.contains` wouldn’t work out of the box without additional implementation.

By using `firstWhereOrNull`, I can avoid the need to override `==` and `hashCode`, and directly match based on a specific property (`name` in this case), which keeps the logic simple and avoids unnecessary changes to the `TransferEncoding` class.

lib/src/headers/headers.dart
```diff
@@ -1237,10 +1235,7 @@ final class _HeadersImpl extends Headers {
 
   /// Apply headers to the response
   @override
-  void applyHeaders(
-    io.HttpResponse response,
-    Body body,
-  ) {
+  void applyHeaders(io.HttpResponse response) {
```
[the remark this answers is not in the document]
klkucaj: Yes, you are right!

lib/src/body/types/mime_type.dart
```diff
@@ -33,6 +33,12 @@ class MimeType {
   /// RTF mime type.
   static const rtf = MimeType('application', 'rtf');
 
+  /// Multipart mime type.
+  static const multipart = MimeType('multipart', 'form-data');
```
SandPod, line 37: Suggestion: Rename to `formData`.

Following the naming used above i would expect this to be named form data? What does dart use?

lib/src/body/types/mime_type.dart
```diff
@@ -33,6 +33,12 @@ class MimeType {
...
+
+  /// Multipart mime type.
+  static const multipartByteranges = MimeType('multipart', 'byteranges');
```
SandPod, line 40: Suggestion: Following how `plainText` is named, this should probably be named `byterangesMultipart` right? What does dart call these?

lib/src/body/types/mime_type.dart
```diff
@@ -74,3 +80,13 @@ extension ContentTypeExtension on ContentType {
   /// We are calling this method 'toMimeType' to avoid conflict with the 'mimeType' property.
   MimeType get toMimeType => MimeType(primaryType, subType);
 }
+
+/// Extension to check if a [MimeType] is not multipart/byteranges.
+extension MimeTypeExtensions on MimeType {
+  /// Checks if the mime type is not multipart/byteranges.
+  bool get isMultipartByteranges {
```
SandPod, line 87: Suggestion: If this is only used in a single place, do the check there directly insteaaed of adding an extention.

lib/src/body/body.dart
```diff
@@ -3,7 +3,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:typed_data';
 
-import 'package:relic/src/body/types/body_type.dart';
+import 'package:relic/relic.dart';
```
SandPod, line 6: Question: This file now only contains a change to this import, is this required?

lib/src/extensions/http_response_extension.dart
```diff
@@ -0,0 +1,68 @@
...
+      httpHeaders.set(key, value);
+    }
+
+    // Set the Content-Type header based on the MIME type of the body.
+    httpHeaders.contentType = body.getContentType();
+
+    // If the content length is known, set it and remove the Transfer-Encoding header.
+    if (body.contentLength != null) {
+      httpHeaders
+        ..contentLength = body.contentLength!
```
SandPod, line 27: Suggestion: Assign the content length before nullcheck so that the dangerous derreff can be avoided ( `!`)

lib/src/extensions/http_response_extension.dart
```diff
@@ -0,0 +1,68 @@
+import 'package:relic/relic.dart';
+import 'dart:io' as io;
+
+import 'package:relic/src/body/types/mime_type.dart';
+
+/// Extension for [io.HttpResponse] to apply headers and body.
+extension HttpResponseExtension on io.HttpResponse {
+  /// Apply headers and body to the response.
+  void applyHeaders(Headers headers, Body body) {
+    var httpHeaders = this.headers;
```
SandPod, line 10: Suggestion: Rename to `responseHeaders` to make it clear what this is.

lib/src/extensions/http_response_extension.dart
```diff
@@ -0,0 +1,68 @@
...
+    httpHeaders.clear();
+
+    // Apply headers
+    var mappedHeaders = headers.toMap();
+    for (var entry in mappedHeaders.entries) {
+      var key = entry.key;
+      var value = entry.value;
```
SandPod, line 17: Suggestion(NIT): Skip assigning this variable and just pass it directly to the set method.

`httpHeaders.set(entry.key, entry.value);`

lib/src/extensions/http_response_extension.dart
```diff
@@ -0,0 +1,68 @@
...
+      // Set Transfer-Encoding header and remove Content-Length as it is not needed for chunked encoding.
+      httpHeaders
+        ..set(
+          Headers.transferEncodingHeader,
+          encodings.map((e) => e.name).toList(),
+        )
+        ..removeAll(Headers.contentLengthHeader);
+    } else {
+      // Set Content-Length to 0 if chunked encoding is not enabled.
+      httpHeaders.contentLength = 0;
```
SandPod, line 65: Question: If we don't use chunked encoding, shouldn't this be the correct content length?

test/relic_server_serve_test.dart
```diff
@@ -508,6 +508,7 @@ void main() {
           return Response.ok(
             body: Body.fromDataStream(
               Stream.value(Uint8List.fromList([1, 2, 3, 4])),
+              contentLength: 4,
```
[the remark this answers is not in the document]
SandPod: No reply and no fix to this comment, Please look at it again.

SandPod (commented):
Couple of more comment and suggestion.

lib/src/extensions/http_response_extension.dart
klkucaj: Your right.

lib/src/body/body.dart
klkucaj: I am going to revert this to main because there no need to do any changes here, the reason why is only this because this is how it is on main https://github.com/klkucaj/relic/blob/c4a9eedd99636336dbf2ce7338b1ccb574d4205b/lib/src/body/body.dart#L6-L7

test/relic_server_serve_test.dart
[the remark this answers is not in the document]
klkucaj: Yes u are right, sorry, i improved other tests and i missed this.

lib/src/body/types/mime_type.dart
klkucaj: @SandPod check the new version of mime type and let me know if its okay.

test/message/body_test.dart
```diff
@@ -0,0 +1,235 @@
...
+      var response = await _get();
+      expect(
+        response.headers['transfer-encoding'],
+        contains(TransferEncoding.chunked.name),
+      );
+      expect(response.body, equals('5\r\nRelic\r\n0\r\n\r\n'));
+    });
+
+    test(
+        'with a valid content length when applying headers '
```
SandPod, line 143: Suggestion: Add a line break for each section in the `Given, When, Then` pattern.

This should be applied to all tests in this file.

For this test that would result in:

```dart
'with a valid content length '
'when applying headers to the response '
'then Content-Length is used instead of chunked encoding'
```

test/message/body_test.dart
```diff
@@ -0,0 +1,235 @@
...
+        Completer<Object?> completer = Completer();
+
+        unawaited(runZonedGuarded(
+          () async {
+            await _scheduleServer(
+              (_) => Response.ok(
+                body: Body.fromDataStream(
+                  Stream.fromIterable([
+                    Uint8List.fromList([1, 2, 3, 4])
+                  ]),
```
SandPod, line 178: Question: What happens if no data is returned from the stream?

lib/src/extensions/http_response_extension.dart
```diff
@@ -0,0 +1,70 @@
...
+    var contentLength = body.contentLength;
+    if (contentLength != null) {
+      responseHeaders
+        ..contentLength = contentLength
+        ..removeAll(Headers.transferEncodingHeader);
+      return;
+    }
+
+    // Check if the content type is multipart/byteranges.
+    var bodyMimeType = body.contentType?.mimeType;
```
SandPod, line 30: Question: What is the difference between `body.contentType` and `body.getContentType`?

lib/src/extensions/http_response_extension.dart
```diff
@@ -0,0 +1,70 @@
...
+        // 304 is not modified
+        statusCode != 304 &&
+        // If the content type is not multipart/byteranges, chunked encoding is applied.
+        !isMultipartByteranges;
+
+    // Prepare transfer encodings.
+    var encodings = headers.transferEncoding?.encodings ?? [];
+    bool isChunked = headers.transferEncoding?.isChunked ?? false;
+
+    if (shouldEnableChunkedEncoding && !isChunked) {
```
SandPod, line 48: Discussion: This still looks a bit funky to me. Let's discuss it.

lib/src/extensions/http_response_extension.dart
klkucaj: Renamed `body.contentType` to `body.bodyType`, this is our implementation of `contentType`.

test/message/body_test.dart
klkucaj: This is removed now, we dont need it anymore.

lib/src/extensions/http_response_extension.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:relic/relic.dart';
+import 'dart:io' as io;
+
+/// Extension for [io.HttpResponse] to apply headers and body.
+extension HttpResponseExtension on io.HttpResponse {
+  /// Apply headers and body to the response.
+  ///
+  /// Resolves conflicts between `Content-Length` and `Transfer-Encoding`
```
SandPod, line 8: Suggestion: Clarify this comment. We no longer resolve conflicts but simply add the the chunked header if required by the body.

SandPod: Combine this with the comment above as well.

lib/src/headers/typed/headers/transfer_encoding_header.dart
```diff
@@ -33,6 +34,11 @@ class TransferEncodingHeader implements TypedHeader {
     return encodings.contains(TransferEncoding.chunked);
   }
 
+  /// Checks if the Transfer-Encoding contains `identity`.
+  bool get isIdentity {
```
SandPod, line 38: Suggestion: Is `isChunked` and `isIdentity` only used in apply headers? If so, these could be extensions there instead of bloating this class.

lib/src/message/message.dart
```diff
@@ -20,19 +20,19 @@ abstract class Message {
     this.context = const {},
   });
 
-  /// Returns the MIME type from the Content-Type header, if available.
-  MimeType? get mimeType => body.contentType?.mimeType;
+  /// Returns the MIME type from the Body-Type (Content-Type) header, if available.
```
SandPod, line 23: Suggestion: Should `header` be inside of the `)` as well?

lib/src/message/response.dart
```diff
@@ -127,6 +125,22 @@ class Response extends Message {
           context: context,
         );
 
+  /// Constructs a 204 No Content response.
+  ///
+  /// This indicates that the request has succeeded but that the server has no
+  /// further information to send in the response body.
+  ///
+  /// {@macro relic_response_body_and_encoding_param}
```
SandPod, line 133: Question: What is the macro here?

test/relic_server_serve_test.dart
```diff
@@ -448,43 +448,28 @@ void main() {
...
-            body: Body.fromDataStream(
-                Stream.value(Uint8List.fromList([1, 2, 3, 4]))),
-          );
-        });
-
-        var response = await _get();
-        expect(response.headers,
-            containsPair(HttpHeaders.transferEncodingHeader, 'chunked'));
-        expect(response.bodyBytes, equals([1, 2, 3, 4]));
+    test('is added when the transfer-encoding header is unset', () async {
```
SandPod, line 451: Suggestion: The description here doesn't really explain the difference in the tests here. The important part is that we don't have a content length in order for the chunked to be added.

SandPod: Same for the other tests in this group.

test/message/body_test.dart
```diff
@@ -0,0 +1,193 @@
...
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:http/http.dart' as http;
+import 'package:relic/relic.dart';
+import 'package:relic/src/method/request_method.dart';
+import 'package:relic/src/relic_server_serve.dart' as relic_server;
+import 'package:test/test.dart';
+
+void main() {
```
SandPod, line 11: Question: Are these the same as the tests in `relic_server_serve_test`?

lib/src/message/response.dart
klkucaj: A way to share comments that are only written in one place.

test/message/body_test.dart
klkucaj: Fixed!

test/relic_server_serve_test.dart
klkucaj: I updated some of the tests, moving a portion to `apply_headers_test.dart`. However, I didn't update the others as they were directly imported from Shelf. I'm unsure if updating these is necessary at this point.
What do u suggest ?

test/message/apply_headers_test.dart
```diff
@@ -0,0 +1,296 @@
...
+
+  @override
+  HttpHeaders get headers => _headers;
+
+  @override
+  int get statusCode => _statusCode;
+}
+
+void main() {
+  group('Given a response', () {
```
SandPod, line 55: Suggestion(NIT): For clarity: `Given a framework response object with some headers when applying headers to http response then headers are applied correctly`?

test/message/apply_headers_test.dart
```diff
@@ -0,0 +1,296 @@
...
+  }) : _statusCode = statusCode;
+
+  @override
+  HttpHeaders get headers => _headers;
+
+  @override
+  int get statusCode => _statusCode;
+}
+
+void main() {
```
SandPod, line 54: Suggestion: We are missing one test here now right, the test when we actually add the transfer chunked encoded? When content-lenght = null.

SandPod (commented):
Last fix!

test/message/apply_headers_test.dart
klkucaj: Done

SandPod (approved):
LGTM! 👍

Issue #18, opened 2024-12-17
There are still places where we use `Object` or `dynamic`


vlidholt:
For instance, `createServer` takes an `Object` for its `address` parameter. This should be typed. We should check if there are other places too, that we have missed.

PR #24, opened 2024-12-23
refactor: Replace Object usage with strongly-typed implementations

klkucaj:
### PR Description
This PR replaces instances of `Object` with strongly-typed implementations to improve code clarity, maintainability, and type safety. 

Key changes include:
- Introducing `RelicAddress` to encapsulate address and port handling.
- Refactoring tests to use the new `createServer` helper for consistent server setup.

No functional changes were made; this is purely a refactor for improved code quality.

Closes: https://github.com/serverpod/relic/issues/18

### DIFF
--- a/lib/src/address/relic_address.dart
+++ b/lib/src/address/relic_address.dart
@@ -1,60 +1,42 @@
 import 'dart:io';
 
 /// A class that represents an address.
-class RelicAddress {
+abstract class RelicAddress<T> {
+  /// Returns the address as an [Object].
+  T get address;
+}
+
+/// A class that represents a hostname address.
+class RelicHostnameAddress implements RelicAddress<String> {
   /// The hostname of the address.
-  final String? hostname;
+  final String hostname;
 
-  /// The internet address of the address.
-  final InternetAddress? internetAddress;
-
-  /// The port of the address.
-  final int port;
-
-  RelicAddress._({
-    this.hostname,
-    this.internetAddress,
-    required this.port,
+  RelicHostnameAddress({
+    required this.hostname,
   });
 
-  /// Creates an address from a string.
-  factory RelicAddress.fromString({
-    required String address,
-    required int port,
-  }) {
-    return RelicAddress._(hostname: address, port: port);
-  }
-
-  /// Creates an address from an [InternetAddress].
-  factory RelicAddress.fromInternetAddress({
-    required InternetAddress address,
-    required int port,
-  }) {
-    return RelicAddress._(internetAddress: address, port: port);
-  }
-
-  /// Returns the address as an [Object].
-  Object get address => hostname ?? internetAddress!;
+  /// Returns the address as a [String].
+  @override
+  String get address => hostname;
 
   @override
-  String toString() {
-    final addr = hostname ?? internetAddress!.address;
-    return '$addr:$port';
-  }
+  String toString() => 'RelicHostnameAddress(hostname: $hostname)';
 }
 
-/// An extension on [InternetAddress] to create a [RelicAddress] with a port.
-extension InternetAddressExtension on InternetAddress {
-  /// Creates a [RelicAddress] with a port.
-  RelicAddress withPort(int port) {
-    return RelicAddress.fromInternetAddress(address: this, port: port);
-  }
-}
+/// A class that represents an internet address.
+class RelicInternetAddress implements RelicAddress<InternetAddress> {
+  /// The internet address of the address.
+  final InternetAddress internetAddress;
 
-/// An extension on [String] to create a [RelicAddress] with a port.
-extension StringExtension on String {
-  /// Creates a [RelicAddress] with a port.
-  RelicAddress withPort(int port) {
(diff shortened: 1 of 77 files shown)


lib/src/address/relic_address.dart
```diff
@@ -0,0 +1,60 @@
...
+  /// Creates an address from an [InternetAddress].
+  factory RelicAddress.fromInternetAddress({
+    required InternetAddress address,
+    required int port,
+  }) {
+    return RelicAddress._(internetAddress: address, port: port);
+  }
+
+  /// Returns the address as an [Object].
+  Object get address => hostname ?? internetAddress!;
```
SandPod, line 37: Suggestion: Make this into two distinct types (Classes) that inherit from RelicAddress.

One that is a wrapper for InternetAddress and one that is a wrapper around a String address.

lib/src/address/relic_address.dart
```diff
@@ -0,0 +1,60 @@
...
+/// A class that represents an address.
+class RelicAddress {
+  /// The hostname of the address.
+  final String? hostname;
+
+  /// The internet address of the address.
+  final InternetAddress? internetAddress;
+
+  /// The port of the address.
+  final int port;
```
SandPod, line 12: Suggestion/Discussion: Remove `port` from the address.

These aren't really connected so could keep them separate.

SandPod: I would expect them to possibly change independently as well.

lib/src/relic_server.dart
```diff
@@ -61,14 +61,14 @@ class RelicServer {
...
+          address.port,
           backlog: backlog,
           shared: shared,
         ),
       false => await HttpServer.bindSecure(
-          address,
-          port,
+          address.address,
+          address.port,
           securityContext!,
```
SandPod, line 72: Is it possible to remove the `!` here?

lib/src/address/relic_address.dart
```diff
@@ -0,0 +1,60 @@
+import 'dart:io';
+
+/// A class that represents an address.
+class RelicAddress {
+  /// The hostname of the address.
+  final String? hostname;
+
+  /// The internet address of the address.
+  final InternetAddress? internetAddress;
```
SandPod, line 9: Question: Should we export this?

test/relic_server_serve_test.dart
```diff
@@ -126,6 +126,7 @@ void main() {
       Uri.http('localhost:$_serverPort', '/'),
     );
     request.sink.add([1, 2, 3, 4]);
+    // ignore: unawaited_futures
     request.sink.close();
```
SandPod, line 130: Suggestion: Can't you wrap this in an `unawaited`?

test/headers/headers_test_utils.dart
```diff
@@ -12,6 +14,25 @@ class BadRequestException implements Exception {
   );
 }
 
+/// Creates a [RelicServer] that listens on the loopback IPv6 address.
+/// If the IPv6 address is not available, it will listen on the loopback IPv4
+/// address.
+Future<RelicServer> createServer({
```
SandPod, line 20: Comment: Nice cleanup 🌟

SandPod (changes requested):
Nice cleanup for the tests. Couple of suggestions to improve the solution and remove some unecessary if, checks.

test/relic_server_serve_test.dart
klkucaj: No, i tried but i got all red https://github.com/serverpod/relic/actions/runs/12580646629/job/35062957328

lib/src/address/relic_address.dart
klkucaj: yes, done

lib/src/relic_server.dart
klkucaj: yes!

lib/src/address/relic_address.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:io';
+
+/// A class that represents an address.
+abstract class RelicAddress<T> {
```
SandPod, line 4: Suggestion: Please add static methods to create the different type of Relic addresses.

Then the users of the interface do not need to know the different types directly.

lib/src/address/relic_address.dart
SandPod: Shouldn't we export `InternetAddress` as well? so that it is enough to import the relic package without having to import the http package?

lib/src/address/relic_address.dart
SandPod: `fromHostname` and `fromInternetAddress`.

lib/src/address/relic_address.dart
klkucaj: yes!

lib/src/address/relic_address.dart
klkucaj: done

SandPod (approved):
LGTM! 👍

PR #29, opened 2025-03-04
refactor!: Get rid of RelicAddress.

nielsenko:
Use InternetAddress from dart:io directly.
Don't allow late or unresolved DNS names anywhere.

## Pre-Launch Checklist

~~- I have referenced at least one issue this PR fixes or is related to.~~
~~- I have added new tests to verify the changes.~~

## Breaking Changes
RelicAddress has been replaced with dart:io InternetAddress


lib/src/relic_server_serve.dart
```diff
@@ -44,7 +44,7 @@ import 'message/response.dart';
 /// {@endtemplate}
 Future<io.HttpServer> serve(
   Handler handler,
-  RelicAddress address,
+  InternetAddress address,
```
SandPod, line 47: Suggestion: Update the documentation with instructions around how to use a simple string to initailze this (maybe also hint towards the loopback?)

SandPod (approved):
Single comment suggestion.

Also wonder if we could remove one of `createServer` or `serve`.

lib/src/relic_server_serve.dart
nielsenko: I feel that information is adequately covered by the docs on the `InternetAddress` class, and will only clutter the docs for serve. Also, as you mention having both `createServer` and `serve` fells wrong, so I plan to revisit this area.

PR #30, opened 2025-03-05
chore: Add pull request title validation.

SandPod:
Adds a simple pull request validation that ensures that we follow conventional commits when creating PRs.

Since we use the pull request title as the squash commit message in main this ensures that we get a consistent commit message style.


nielsenko (approved):
LGTM!

PR #36, opened 2025-03-13
ci: Add test coverage

nielsenko:
Add test coverage reporting to track code execution during tests and identify
untested areas. Implement additional quality checks:

- Publish dry-run to verify package meets publishing requirements
- Pana score analysis to evaluate package quality and documentation
- Dependency bounds checking to ensure proper version constraints

## Summary by CodeRabbit

- **New Features**
	- Introduced a consolidated build process that verifies dependency integrity, formatting, analysis, and performs a publishing dry run.
	- Expanded testing with enhanced support across multiple Dart SDK versions.

- **Chores**
	- Updated dependency versions and incorporated new packages to optimize package management.

### DIFF
--- a/.github/workflows/dart-tests.yaml
+++ b/.github/workflows/dart-tests.yaml
@@ -64,7 +64,6 @@ jobs:
 
   unit_tests:
     name: Run Unit Tests
-    needs: build
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/dart-tests.yaml
+++ b/.github/workflows/dart-tests.yaml
@@ -1,135 +1,104 @@
-name: Relic CI
+name: Relic Ci
 
 on:
   push:
-    branches:
-      - main
-      - dev
+    branches: [main]
   pull_request:
-    branches:
-      - main
-      - dev
+    branches: [main]
+  schedule:
+    - cron: "0 0 * * 0" # every Sunday at 00:00 UTC
+  workflow_dispatch: # allow manual triggering
 
 env:
   PUB_CACHE_PATH: ~/.pub-cache
 
 jobs:
-  dart_format:
-    name: Verify Dart Formatting
+  build:
+    name: Verify Build
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
-        dart_sdk: ['3.3.0']
+        os: [ubuntu-latest]
+        dart_sdk: ["3.3.0", stable]
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Dart
         uses: dart-lang/setup-dart@v1.3
         with:
           sdk: ${{ matrix.dart_sdk }}
-      - name: Cache Dart dependencies
-        uses: actions/cache@v3
+
+      - name: Cache Dart Dependencies
+        uses: actions/cache@v4
         with:
           path: ${{ env.PUB_CACHE_PATH }}
           key: ${{ runner.os }}-pub-cache-${{ matrix.dart_sdk }}
-          restore-keys: |
-            ${{ runner.os }}-pub-cache-
-      - name: Verify formatting
+          restore-keys: ${{ runner.os }}-pub-cache-
+
+      - name: Verify Formatting
+        # Disable formatting check on stable for now.
+        # Dart stable (3.7.2 as of 2025-03-14) has radically different formating
+        if: ${{ matrix.dart_sdk != 'stable' }}
         run: dart format --output=none --set-exit-if-changed .
 
-  dart_analyze:
-    name: Run Dart Analysis
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
-        dart_sdk: ['3.3.0', 'stable']
-    steps:
-      - name: Checkout Code
-        uses: actions/checkout@v3
-      - name: Setup Dart
-        uses: dart-lang/setup-dart@v1.3
-        with:
-          sdk: ${{ matrix.dart_sdk }}
-      - name: Cache Dart dependencies
-        uses: actions/cache@v3
-        with:
-          path: ${{ env.PUB_CACHE_PATH }}
-          key: ${{ runner.os }}-pub-cache-${{ matrix.dart_sdk }}
-          restore-keys: |
-            ${{ runner.os }}-pub-cache-
-      - name: Install dependencies
-        run: dart pub upgrade
-      - name: Analyze
-        run: dart analyze
+      - name: Check Dependency Bounds
+        if: ${{ matrix.dart_sdk != '3.3.0' }} # --tighten available from 3.5
+        run: >-
+          dart pub downgrade --tighten &&
+          dart pub upgrade --major-versions &&
+          git diff-files --quiet
 
-  dart_analyze_downgrade:
-    name: Run Dart Analysis Downgrade
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout Code
-        uses: actions/checkout@v3
-      - name: Setup Dart
-        uses: dart-lang/setup-dart@v1.3
-        with:
-          sdk: '3.3.0'
-      - name: Cache Dart dependencies
-        uses: actions/cache@v3
-        with:
-          path: ${{ env.PUB_CACHE_PATH }}
-          key: ${{ runner.os }}-pub-cache-downgrade
-          restore-keys: |
-            ${{ runner.os }}-pub-cache-
-      - name: Install dependencies (Downgrade)
-        run: dart pub downgrade
       - name: Analyze
-        run: dart analyze
+        run: >-
+          dart pub get &&
+          dart analyze --fatal-infos
 
-  dart_analyze_latest_downgrade:
-    name: Run Dart Analysis Latest (stable) Downgrade
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout Code
-        uses: actions/checkout@v3
-      - name: Setup Dart
-        uses: dart-lang/setup-dart@v1.3
-        with:
-          sdk: 'stable'
-      - name: Cache Dart dependencies
-        uses: actions/cache@v3
-        with:
-          path: ${{ env.PUB_CACHE_PATH }}
-          key: ${{ runner.os }}-pub-cache-downgrade
-          restore-keys: |
-            ${{ runner.os }}-pub-cache-
-      - name: Install dependencies (Downgrade)
-        run: dart pub downgrade
-      - name: Analyze
-        run: dart analyze
+      - name: Check Pana Score
+        if: ${{ matrix.dart_sdk == 'stable' }}
+        run: >-
+          dart pub global activate pana &&
+          dart pub global run pana --no-warning --exit-code-threshold=0 --json
+
+      - name: Publish Dry Run
+        run: dart pub publish --dry-run
 
   unit_tests:
     name: Run Unit Tests
+    needs: build
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        dart_sdk: ['3.3.0']
+        dart_sdk: ["3.3.0", stable, beta]
+        deps: [downgrade, upgrade]
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Dart
         uses: dart-lang/setup-dart@v1.3
         with:
           sdk: ${{ matrix.dart_sdk }}
-      - name: Cache Dart dependencies
-        uses: actions/cache@v3
+
+      - name: Cache Dart Dependencies
+        uses: actions/cache@v4
         with:
           path: ${{ env.PUB_CACHE_PATH }}
           key: ${{ runner.os }}-pub-cache-${{ matrix.dart_sdk }}
-          restore-keys: |
-            ${{ runner.os }}-pub-cache-
-      - name: Install dependencies
-        run: dart pub upgrade
-      - name: Run tests
-        run: dart test
+          restore-keys: ${{ runner.os }}-pub-cache-
+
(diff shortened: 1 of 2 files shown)


nielsenko:
@coderabbitai pause

nielsenko:
@coderabbitai summary

nielsenko:
@coderabbitai configuration

nielsenko:
@coderabbitai full

nielsenko:
@coderabbitai resolve

pubspec.yaml
```diff
@@ -4,21 +4,28 @@ version: 0.3.0
...
+  test: ^1.25.0
   test_descriptor: ^2.0.1
   mockito: ^5.4.4
+  # The following are not used directly, but are included transitively.
+  # Due to bad semver hygiene we need to bound the versions higher than
+  # our direct dependencies indicate.
+  file: ^7.0.0
+  frontend_server_client: ^4.0.0
+  pub_semver: ^2.1.4
+  watcher: ^1.1.0
```
nielsenko, line 31: As we are now actually testing on the lower-bounds as well, we need to accommodate for other packages bad semver hygiene.

To be honest it is very easy mistake to make in Dart, as just adding a new methods on a class is a breaking change (since someone might implement the class).

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
-      - name: Install dependencies
-        run: dart pub upgrade
-      - name: Analyze
-        run: dart analyze
+      - name: Check Dependency Bounds
+        if: ${{ matrix.dart_sdk != '3.3.0' }} # --tighten available from 3.5
+        run: >-
+          dart pub downgrade --tighten &&
+          dart pub upgrade --major-versions &&
+          git diff-files --quiet
```
nielsenko, line 49: This helps us catch if dependencies are released in new major versions, and to see if our lower constraints are artificially low.

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
+        run: >-
+          dart pub global activate pana &&
+          dart pub global run pana --no-warning --exit-code-threshold=0 --json
+
+      - name: Publish Dry Run
+        run: dart pub publish --dry-run
 
   unit_tests:
     name: Run Unit Tests
+    needs: build
```
nielsenko, line 67: Since I spawning an awful lot of runners in this job, I make sure only to do it if the fast and simply build job completes. For the sake of the planet and all..

pubspec.yaml
```diff
@@ -4,21 +4,28 @@ version: 0.3.0
...
-  sdk: '>=3.3.0 <4.0.0'
-  
+  sdk: ">=3.3.0 <4.0.0"
+
 dependencies:
   async: ^2.11.0
   collection: ^1.17.1
   http_parser: ^4.0.2
-  path: ^1.8.2
+  path: ^1.8.3
```
nielsenko, line 13: Bumped due to `Check Dependency Bounds`

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
-        os: [windows-latest, ubuntu-latest, macos-latest]
-        dart_sdk: ['3.3.0']
+        os: [ubuntu-latest]
+        dart_sdk: ["3.3.0", stable]
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Dart
         uses: dart-lang/setup-dart@v1.3
```
nielsenko, line 27: Using dart-lang/setup-dart is nice and fast. One problem is that most people will install Dart with Flutter, and the version of Dart stable and Dart from Flutter stable often differs. Fx. Dart stable is 3.7.2 as of today, but Dart from Flutter stable is currently 3.7.0 (and likely won't be bumped, until they go straight to 3.8.0).

Perhaps it is better to install Flutter stable and use that version of Dart?

.github/workflows/dart-tests.yaml
SandPod, line 49: Question: Don't we need to run an analyze after both the `downgrade` and the `upgrade` to catch if there are any issues with the dependencies?

.github/workflows/dart-tests.yaml
SandPod: Hmm, Don't we want as much information as possible in one run in the CI pipe? Seems like these could be broken for different reasons (for example som fatal info), but the feedback is still valuable?

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
-          key: ${{ runner.os }}-pub-cache-downgrade
-          restore-keys: |
-            ${{ runner.os }}-pub-cache-
-      - name: Install dependencies (Downgrade)
-        run: dart pub downgrade
       - name: Analyze
-        run: dart analyze
+        run: >-
+          dart pub get &&
+          dart analyze --fatal-infos
```
SandPod, line 54: Question: Should this be ignored for `stable`?

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
   unit_tests:
     name: Run Unit Tests
+    needs: build
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        dart_sdk: ['3.3.0']
+        dart_sdk: ["3.3.0", stable, beta]
+        deps: [downgrade, upgrade]
```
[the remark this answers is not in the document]
SandPod: Question: Theoretically, If dart releases a bad version into beta this will prevent us from merging PRs. Do we see this as highly unlikely or should we exclude the beta from a required pass?

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
+        continue-on-error: ${{ matrix.dart_sdk != 'stable' }}
+        run: >-
+          dart pub global activate coverage &&
+          dart pub global run coverage:test_with_coverage --branch-coverage
+
+      - name: Upload Coverage
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: coverage/lcov.info
```
SandPod, line 104: Question: Does codecov publish the report as part of the PR in some way?

SandPod (commented):
Neat cleanup! 👍 

I have a couple of questions!

.github/workflows/dart-tests.yaml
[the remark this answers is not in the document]
SandPod: Perhaps we should have the same thinking with Stable. If Dart releases a new version we are not compatible with that should not block our PR flow. It could be a job we run in main.

.github/workflows/dart-tests.yaml
nielsenko: I decided not to, but I can definitely add it back.

Analysis is only looking at our code, not the dependencies, and as such it should not be impacted by upgrade/downgrade as far as I can tell.

As long as we can resolve the constraints in both cases.

However, running the actual tests will often reveal that the constraints are not correct.

.github/workflows/dart-tests.yaml
nielsenko: Yes, it will be available from the PR, once we have a baseline.

You can also drill into the step and find an url already

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
+      - name: Install Dependencies (${{ matrix.deps }})
+        run: >-
+          dart pub ${{ matrix.deps }} &&
+          dart pub get
+
+      - name: Run Tests With Coverage
+        continue-on-error: ${{ matrix.dart_sdk != 'stable' }}
+        run: >-
+          dart pub global activate coverage &&
+          dart pub global run coverage:test_with_coverage --branch-coverage
```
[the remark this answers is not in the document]
nielsenko: For the integration tests you run the test-suite as above, but also make sure the server is started in a similar way, so you can catch coverage from both client and server.

I cannot recall how to do it of the top of my head, but as I recall `coverage:test_with_coverage` can be split into multiple steps, and for the server-side you would need some of those.

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
 
   unit_tests:
     name: Run Unit Tests
+    needs: build
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        dart_sdk: ['3.3.0']
+        dart_sdk: ["3.3.0", stable, beta]
```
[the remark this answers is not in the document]
nielsenko: Notice the `continue-on-error: ${{ matrix.dart_sdk != 'stable' }}`. I think it is nice to get a heads-up on upcoming issues like this

.github/workflows/dart-tests.yaml
```diff
@@ -1,135 +1,104 @@
...
+    - cron: "0 0 * * 0" # every Sunday at 00:00 UTC
+  workflow_dispatch: # allow manual triggering
 
 env:
   PUB_CACHE_PATH: ~/.pub-cache
 
 jobs:
-  dart_format:
-    name: Verify Dart Formatting
+  build:
```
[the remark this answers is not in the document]
nielsenko: A job will require a runner, which you will have to wait for, and it has a certain startup time. 

I prefer to keep things as steps unless there is a clear benefit to running stuff in parallel.

I would like to know more about how the analyzer would come to different conclusions on windows. The reason to run only on linux is that I didn't deem that possible.

.github/workflows/dart-tests.yaml
nielsenko: Analysis **can** have different results per version. We can definitely end in a situation where we need to _choose_ what version to comply to, but for know that is not the case with relic.

.github/workflows/dart-tests.yaml
nielsenko: BTW. Not sure why I cannot reply to the question just above:

> Question: Theoretically, If dart releases a bad version into beta this will prevent us from merging PRs. Do we see this as highly unlikely or should we exclude the beta from a required pass?

We don't block on a failure on `beta`, only on `stable`.
```yaml
continue-on-error: ${{ matrix.dart_sdk != 'stable' }}
```

.github/workflows/dart-tests.yaml
nielsenko: The first job is fast.

Also, to be honest my plan is to make sure caches are primed even more in this job, so that the latter job can benefit. Doing that mandates the `needs` to ensure the caches are actually primed.

.github/workflows/dart-tests.yaml
SandPod: Since we are a backend framework that I think it makes sense to focus directly on dart. Most cases might go through Flutter, but out target is simply Dart developers.

One deploy case might be a container that simply has Dart inside.

.github/workflows/dart-tests.yaml
SandPod: I think a breaking change in the API of a dependency will not be visible when running the constraints checks. We might still be able to resolve the dependency, but we are using it in an invalid way. 

And since unit tests run in isolation, we would only detect it if we have 100% code coverage for all paths in our code which is an assumption I don't think we should have.

So the analyzer would be needed to catch these.

.github/workflows/dart-tests.yaml
SandPod: We might as well codify the intended behavior here. The lowest bound will need to be the marker that we want to 100% comply towards.

SandPod: We still might want `--fatal-warnings` for stable though.

.github/workflows/dart-tests.yaml
SandPod: > The first job is fast.

True, but pushing a new commit only to discover that there is a failure in the unit tests that was not caught is slow.

I think we should optimize developer time here.

.github/workflows/dart-tests.yaml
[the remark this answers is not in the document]
SandPod: What happens if we fail with a `continue-on-error`? How are you notified that it failed? Or is the job marked as successfull anyways and you need to read the logs to understand if it failed?

.github/workflows/dart-tests.yaml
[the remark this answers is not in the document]
SandPod: I really think we should optimize developer roundtrip time here. Pushing a change only to resolve one step and having to wait for the CI pipe again is a pretty slow developer experience. we want to give as much information in one go as possible.

.github/workflows/dart-tests.yaml
nielsenko: I'll add it back, but I'm not sure why you think the analyzer will catch this situation?

.github/workflows/dart-tests.yaml
nielsenko: Okay, I'll keep it as is.

.github/workflows/dart-tests.yaml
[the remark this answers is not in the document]
nielsenko: The job still fails (red cross), it just doesn't break the pipeline. It will be very visible when you are about to merge.

.github/workflows/dart-tests.yaml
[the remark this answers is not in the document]
nielsenko: Okay. Only lowest sdk (ie. 3.3.0) will break the build.

.github/workflows/dart-tests.yaml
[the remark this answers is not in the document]
SandPod: Windows - that was an issue in combination with the code generator. We accidentally generated platform specific import paths which was not good... 🙃

.github/workflows/dart-tests.yaml
SandPod: The analyzer reports invalid dart code right?

SandPod (approved):
LGTM! 👍 

I still think it might be better to have more pipes running in parallel to mininize number of roundtrips through the CI. But we could break that out if it becomes an issue.

PR #37, opened 2025-03-15
ci: Hoist continue-on-error to job

nielsenko:
Move continue-on-error clause to job instead of step.

If continue-on-error is set to true on a job, then a job won't fail the workflow, but will still be marked red, on failure.
This is not the case for steps.

Also ensures we upload coverage even if there are test failures.

## Summary by CodeRabbit

- **Chores**
	- Updated internal workflows to enhance testing consistency and coverage reporting.
	- Improved conditions for error handling in test execution.

### DIFF
--- a/.github/workflows/dart-tests.yaml
+++ b/.github/workflows/dart-tests.yaml
@@ -21,11 +21,14 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         dart_sdk:
-          - "3.3.0" # ${{ env.LOWEST_DART_SDK }} won't work ¯\_(ツ)_/¯
+          # ${{ env.LOWEST_DART_SDK }} won't work at job level as env context not available for strategy ¯\_(ツ)_/¯
+          # (see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#context-availability)
+          - "3.3.0"
           - stable
     steps:
       - name: Checkout Code
         uses: actions/checkout@v4
+
       - name: Setup Dart
         uses: dart-lang/setup-dart@v1.3
         with:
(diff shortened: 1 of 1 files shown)


nielsenko:
This is to rectify a false claim I made previously: https://github.com/serverpod/relic/pull/36#discussion_r1995463435 

_continue_on_error_ behaviour differs on steps vs. jobs

nielsenko:
@coderabbitai review

nielsenko:
@coderabbitai full

nielsenko:
@SandPod While working on this I learned that you cannot use the _env_ context outside of a step. I'm not sure I understand/agree with the _why_, but it is what it is.

We could consider moving the `LOWEST_SDK_VERSION` from _env_ to _vars_ which is available, but then we have to set it as variable on the repo (which is an admin op). However, If we do that we can get rid of hardcoded "3.3.0" and most of the comments.

.github/workflows/dart-tests.yaml
```diff
@@ -101,13 +106,13 @@ jobs:
...
 
       - name: Run Tests With Coverage
-        continue-on-error: ${{ matrix.dart_sdk != env.LOWEST_DART_SDK }}
         run: >-
           dart pub global activate coverage &&
           dart pub global run coverage:test_with_coverage --branch-coverage
 
       - name: Upload Coverage
         uses: codecov/codecov-action@v5
+        if: ${{ !cancelled() }} # upload coverage irrespective of test results
```
SandPod, line 115: Question(For my knowledge): Where is this `cancelled()` function defined? Is it part of github actions?

SandPod (approved):
🚀

SandPod:
> @SandPod While working on this I learned that you cannot use the _env_ context outside of a step. I'm not sure I understand/agree with the _why_, but it is what it is.
> 
> We could consider moving the `LOWEST_SDK_VERSION` from _env_ to _vars_ which is available, but then we have to set it as variable on the repo (which is an admin op). However, If we do that we can get rid of hardcoded "3.3.0" and most of the comments.

We could add it there if you would like, but It might be a bit implicit to the running pipe? Might be better to have it hardcoded in the action.

.github/workflows/dart-tests.yaml
nielsenko: Yes. You have `success()` , `failure()`, and `cancelled()` to query the state of the latest step. `!cancelled()` is just short for `success() || failure()`.

There is also `always()`, but I don't want to upload on _cancelled_, so I don't use that

SandPod: Cool! 👍

Issue #35, opened 2025-03-12
Refactor lazy parsing


nielsenko:
Header parsing and serialization should have minimal overhead compared to raw string operations. This means avoiding unnecessary object allocations, optimizing parsing algorithms, and ensuring that typed headers don't become a bottleneck in request processing.

The current implementation pays for typed headers up front in terms of memory usage. Some effort has gone into supporting lazy parsing, but some operations still force eager parsing.

Issue #33, opened 2025-03-12
Refactoring Typed Headers


nielsenko:
The current implementation of typed headers presents several challenges that need addressing:

- Flexibility & customization
  Developers should be able to define their own typed header classes, customize parsing behavior, and extend functionality as needed without modifying the core library. It should be possible to mix-and-match strict vs non-strict parsing modes on a per-header, per-handler basis.

- Opt-in approach
  Forcing typed headers on all users create unnecessary complexity for some use cases. A gradual adoption approach allows developers to migrate at their own pace, maintaining backward compatibility with existing code while providing a clear path forward.

- Efficiency & performance
  Header parsing and serialization should have minimal overhead compared to raw string operations. This means avoiding unnecessary object allocations, optimizing parsing algorithms, and ensuring that typed headers don't become a bottleneck in request processing.

  The current implementation pays for typed headers up front in terms of memory usage. Some effort has gone into supporting lazy parsing, but some operations still force eager parsing.

- Custom header support
  Beyond standard HTTP headers, applications often define custom headers for specific functionality. The typed header system should provide an extensible framework that allows developers to easily define and use their own custom header types with the same benefits as built-in headers.

- Inter-isolate sharing
  Since Dart applications often use multiple isolates for concurrency, header objects should be easily transferable between isolates without expensive serialization/deserialization cycles. This requires careful consideration of the internal representation of header data.

PR #38, opened 2025-03-20
feat!: Support typed access to custom headers

nielsenko:
## Description

_All headers are created equal_

This PR allows any header to be typed. We no longer distinguishes between standard and custom headers in any way.
Both can be handled as raw or typed values.

The PR also reduces the overhead associated with headers that are never used. Previously instances of the `Headers` class would for each standard header keep track of a `T?`, a `bool`, and a reference to an initializer function for parsing, even if none of these where ever used. The parsing would happen lazy, but the memory cost was paid up front.

Instead typed headers are handled using const constructed flyweights objects (`HeaderAccessor<T>`) that avoids duplicated initializer references, and cache parsed values on the `Headers` instance using `Expando` objects.

## Related Issues
- Resolves: #9 
- Resolves: #11 
- Resolves: #33 
- Resolves: #35 
- Resolves: #41 

## Pre-Launch Checklist

- [ ] This update focuses on a single feature or bug fix. (For multiple fixes, please submit separate PRs.)
- [X] I have read and followed the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style) and formatted the code using [dart format](https://dart.dev/tools/dart-format).
- [X] I have referenced at least one issue this PR fixes or is related to.
- [x] All existing and new tests pass successfully. (But some skipped)

## Breaking Changes
- [X] Includes breaking changes.
- [ ] No breaking changes.

The public api of the `Headers` class changes.

## Additional Notes
_Include any additional information or context about the PR here, if needed._

## Summary by CodeRabbit

- **Improvements**
  - Consolidated HTTP header management improves consistency and reliability in request–response handling.
  - Enhanced validation and error messaging deliver more predictable interactions.
  - New functionality for managing headers, including better handling of exceptions and streamlined header processing.

- **Bug Fixes**
  - Corrected header naming and parsing issues for more robust behavior.
  - Improved error handling for invalid headers ensures exceptions are thrown as expected.

- **Tests & Maintenance**
  - Extensive test suite refinements ensure robust header validation.
  - Minor dependency updates enhance the overall development environment.
  - Introduced new test cases and parameters for more precise header manipulation during tests.


test/headers/headers_flyweight_test.dart
```diff
@@ -0,0 +1,301 @@
...
+  String toHeaderString() => _value;
+}
+
+void main() {
+  group('HeaderDecoderSingle', () {
+    late HeaderDecoderSingle<int> decoder;
+
+    setUp(() {
+      decoder = HeaderDecoderSingle<int>((s) => int.parse(s));
+    });
```
SandPod, line 23: Question: Do these really need to be in a setup? We are just constructing the object right? We are not actually invoking it until the tests.

SandPod: Same for all decoders.

test/headers/headers_flyweight_test.dart
```diff
@@ -0,0 +1,301 @@
...
+      var headers = Headers.build((mh) => mh['x-test'] = ['123']);
+
+      // Access value twice to test caching
+      var value1 = flyweight.getValueFrom(headers);
+
+      // Change the header value - if caching works, this won't affect the second getValue
+      var headers2 = headers.transform((mh) {
+        expect(flyweight.getValueFrom(mh), equals(123));
+        mh['x-test'] = ['456'];
+        // expect(flyweight.getValueFrom(mh), equals(456)); // <-- THIS DON' CURRENTLY WORK!!
```
SandPod, line 92: Question: Why does this not work?

test/headers/headers_flyweight_test.dart
```diff
@@ -0,0 +1,301 @@
...
+      });
+      var value2 = flyweight.getValueFrom(headers);
```
SandPod, line 94: Question: We are reading the value from the same headers again. I thought `transform` was basically copy with, I would not expect transform to modify `headers`, but my changes should be contained in `headers2`.

test/headers/headers_flyweight_test.dart
```diff
@@ -0,0 +1,301 @@
...
+
+      expect(value1, equals(123));
+      expect(value2, equals(123)); // Should be from cache, not the new value
+
+      // Create a new headers object to test a fresh value
+      var value3 = flyweight.getValueFrom(headers2);
+
+      expect(value3, equals(456)); // Should get the new value
```
SandPod, line 102: Suggestion: This seems like a different test/scenario.

Please brake it out.

test/relic_server_serve_test.dart
```diff
@@ -101,11 +103,13 @@ void main() {
...
-          isNot(contains(HttpHeaders.transferEncodingHeader)),
-        );
-
+        // TODO(nielsenko): Enforcing this absense Transfer-Encoding will cause
+        // trouble with 'transfer_encoding_header_test.dart'
+        //
+        // expect(
+        //   request.headers,
+        //   isNot(contains(HttpHeaders.transferEncodingHeader)),
+        // );
```
SandPod, line 112: Question: Why is this commented out?

lib/src/headers/header_flyweight.dart
```diff
@@ -0,0 +1,162 @@
...
+import 'package:relic/src/headers/typed/typed_header_interface.dart';
+
+import 'exception/invalid_header_exception.dart';
+import 'headers.dart';
+
+/// Flyweight class for headers. It should always be const constructed.
+///
+/// The externalized state is stored in the Headers external, and to avoid repeated
+/// parsing of the raw value the value is cached with an [Expando].
+final class HeaderFlyweight<T extends Object> {
```
SandPod, line 13: Discussion: Let's walk through this implementation in a pair review session.

test/headers/basic/access_control_allow_credentials_header_test.dart
```diff
@@ -15,118 +15,124 @@ import '../docs/strict_validation_docs.dart';
 
 void main() {
   group(
-      'Given an Access-Control-Allow-Credentials header with the strict flag true',
-      () {
-    late RelicServer server;
+    'Given an Access-Control-Allow-Credentials header with the strict flag true',
```
SandPod, line 18: Question: How can the strict mode tests still pass? I can't see us introducing the middleware?

SandPod (commented):
Nice work!

Had a bunch of questions, mostly related to me not fully understanding the flyweight pattern. So maybe we could have a call and walk through that?

Otherwise I found some cleanup that was needed.

lib/src/headers/header_flyweight.dart
```diff
@@ -0,0 +1,162 @@
...
+  const HeaderFlyweight(this.key, this.decode);
+
+  Iterable<String>? rawFrom(HeadersBase external) => external[key];
+
+  V getValueFrom<V extends T?>(
+    HeadersBase external, {
+    V Function(Exception) orElse = _raise,
+  }) {
+    try {
+      try {
```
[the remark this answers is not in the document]
nielsenko: The inner is for `as V` failing when V not nullable and `raw == null`. The outer is for if `orElse` throws, in which case we need to wrap the exception in a `InvalidHeaderException`.

lib/src/headers/header_flyweight.dart
```diff
@@ -0,0 +1,162 @@
...
+  Expando get _cache => _caches.putIfAbsent(this, Expando.new);
+
+  final String key;
+  final HeaderDecoder<T> decode;
+
+  const HeaderFlyweight(this.key, this.decode);
+
+  Iterable<String>? rawFrom(HeadersBase external) => external[key];
+
+  V getValueFrom<V extends T?>(
```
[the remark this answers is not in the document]
nielsenko: Yes it should. I'll check up on the linter rule.

lib/src/headers/header_flyweight.dart
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/// A "class" representing a typed header.
+///
+/// Instances are intended to be short-lived. They are typically used as
+/// temporary objects during header processing.
+///
+/// This is implemented as an extension type over a record type
+/// to keep runtime cost low.
+// ignore: library_private_types_in_public_api
```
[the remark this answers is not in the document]
nielsenko: We are ignoring the private typedef `_HeaderTuple`. Since it is a record type it has structural equality, and you can just pass the record, ie: `(flyweight: x, headers: y)`. It is just for internal convenience.

lib/src/headers/headers.dart
```diff
@@ -90,1830 +95,60 @@ abstract base class Headers {
...
-        _contentLanguage = contentLanguage,
-        _contentLocation = contentLocation,
-        _contentRange = contentRange,
-        _etag = etag,
-        _proxyAuthenticate = proxyAuthenticate,
-        _retryAfter = retryAfter,
-        _trailer = trailer,
-        _vary = vary,
-        _wwwAuthenticate = wwwAuthenticate,
+  // cannot be made const before: <insert link to dart CL
```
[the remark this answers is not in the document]
nielsenko: I would prefer to tag this as a TODO, and add the link I'm missing. How about that?

lib/src/message/response.dart
```diff
@@ -287,14 +287,22 @@ class Response extends Message {
...
     }
   }
 
+  static Headers _ensureDate(Headers? headers) {
+    headers ??= Headers.empty();
+    if (!headers.date_.isSet) {
+      headers = headers.transform((mh) => mh.date ??= DateTime.now());
+    }
+    return headers;
+  }
```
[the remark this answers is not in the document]
nielsenko: It is not really required, but `headers.transform` will always produce a new `Headers` object, even if `mh.date` is already set, and `mh.date ??= ..` is a no-op. This is not a probllem per-se, but we have some tests that use `same` matcher.

lib/src/headers/standard_headers_flyweights.dart
```diff
@@ -0,0 +1,351 @@
+import 'package:relic/src/method/request_method.dart';
+
+import 'header_flyweight.dart';
+import 'headers.dart';
+import 'parser/common_types_parser.dart';
+import 'typed/typed_headers.dart';
+
+/// Date-related headers
+const date = HeaderFlyweight<DateTime>(
```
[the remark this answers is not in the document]
nielsenko: It is intended. Think of them like providers in riverpod. We could have them as static members on `Headers`(perhaps replacing the header name strings) thereby "name-spacing" them. A reason (not sure if it is good enough) for not doing that is that it again makes some headers lesser than others.

Since these are all in a separate file, you can namespace them on import.

lib/src/headers/parser/common_types_parser.dart
```diff
@@ -131,53 +128,21 @@ String parseString(String value) {
...
 ///
 /// - Throws a [FormatException] if the resulting list is empty.
-List<String> parseStringList(List<String> values) {
+List<String> parseStringList(Iterable<String> values) {
   var tempValues = values.splitTrimAndFilterUnique();
   if (tempValues.isEmpty) {
     throw FormatException('Value cannot be empty');
   }
-  return tempValues;
+  return tempValues.toList();
```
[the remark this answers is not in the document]
nielsenko: Maybe, I did consider it. So far I have only used `Iterable<T>` on the raw values, not the typed ones.

lib/src/headers/typed/headers/accept_header.dart
```diff
@@ -56,8 +56,8 @@ class MediaRange {
   /// This method processes the media range string, extracting the type,
   /// subtype, quality, and parameters.
   factory MediaRange.parse(String value) {
-    final parts = value.splitTrimAndFilterUnique(separator: ';');
-    final typeSubtype = parts[0].split('/');
+    final parts = value.splitTrimAndFilterUnique(separator: ';').toList();
```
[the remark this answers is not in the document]
nielsenko: :thinking:

test/headers/basic/access_control_allow_credentials_header_test.dart
nielsenko: Maybe I misunderstood our earlier discussions, but I thought we decided to introduce that later.

My idea would be to replace `touchHeaders` with a middleware that validate those headers, something like `strictValidation([date, from])`, by basically _touching_ all those headers (flyweights).

test/headers/headers_flyweight_test.dart
```diff
@@ -0,0 +1,301 @@
...
+  final String _value;
+
+  TestTypedHeader(this._value);
+
+  @override
+  String toHeaderString() => _value;
+}
+
+void main() {
+  group('HeaderDecoderSingle', () {
```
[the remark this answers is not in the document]
nielsenko: Okay, I will.

test/relic_server_serve_test.dart
nielsenko: This is the `isNot(contains(HttpHeaders.transferEncodingHeader))` that accidentally passed before when `Headers` was not a collection.

We no longer strip `Transfer Encoding` per the stale copy-paste requirement in the `README`. Also they way it is stripped in shelf is way too simple.

lib/src/message/response.dart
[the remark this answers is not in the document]
SandPod: Move this to the adapter layer (RelicServer)

lib/src/headers/standard_headers_flyweights.dart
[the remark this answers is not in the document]
SandPod: Moves these to `Headers`.

lib/src/headers/header_flyweight.dart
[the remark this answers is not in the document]
nielsenko: I'll make `HeaderTuple` typedef public and remove the ignore

lib/src/headers/standard_headers_flyweights.dart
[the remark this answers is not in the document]
nielsenko: These will be made const static members on the `Headers` class. We will the const static `<x>Header` strings  from the same, as these can the be accessed via the flyweight, ie `Headers.date.key`.

This also gets rid of `standard_headers_flyweight.dart`. We will also make `standers_headers_extension.dart` a part of `headers.dart`.

test/relic_server_serve_test.dart
nielsenko: Removed stale test

lib/src/headers/headers.dart
```diff
@@ -1,14 +1,19 @@
-import 'dart:io' as io;
+import 'dart:collection';
+import 'dart:io' as io; // TODO: Get rid of this dependen
```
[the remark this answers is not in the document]
SandPod: As long as we clear these out in the next PR and never have an "incomplete" state in main I'm happy.

lib/src/headers/headers.dart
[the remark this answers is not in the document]
SandPod: Yes, that is OK! As long as it is actionable by anyone.

lib/src/headers/header_flyweight.dart
```diff
@@ -0,0 +1,199 @@
...
+  required String key,
+}) {
+  var message = "$exception";
+  if (exception is InvalidHeaderException) {
+    message = exception.description;
+  } else if (exception is FormatException) {
+    message = exception.message;
+  } else if (exception is ArgumentError) {
+    message = exception.message;
+  }
```
SandPod, line 179: Question: CodeCov warns that these lines are not tested, should we act on them?

lib/src/headers/standard_headers_extensions.dart
```diff
@@ -0,0 +1,281 @@
+import 'package:relic/src/method/request_method.dart';
+import 'headers.dart';
+import 'typed/typed_headers.dart';
+
+extension HeadersEx on Headers {
+  DateTime? get date => Headers.date[this]();
```
SandPod, line 6: Question(For my knowledge): Shouldn't these collide with the static const variable in `headers.dart`?

SandPod: The comment also mentioned that these would not be part of `headers.dart`, what you mean by that?

test/headers/headers_flyweight_test.dart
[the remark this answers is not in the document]
SandPod: This has not been applied.

test/headers/headers_flyweight_test.dart
SandPod: This has not been applied.

test/headers/headers_flyweight_test.dart
SandPod: Explained in call.

test/headers/headers_flyweight_test.dart
SandPod: This has not been applied.

test/headers/header_test.dart
```diff
@@ -1,152 +1,458 @@
...
-              'Unexpected additional headers: $unexpectedAdditionalHeaders.',
-        );
+      singleTest('then raw value not set', headers, isNot(contains(v.key)));
+      singleTest('then isSet is false', header.isSet, isFalse);
+      singleTest('then isValid is false', header.isValid, isFalse);
+      singleTest('then valueOrNull is null', header.valueOrNull, isNull);
+      singleTest('then valueOrNullIfInvalid is null',
+          header.valueOrNullIfInvalid, isNull);
+      singleTest('then value throws', () => header.value, throwsMissingHeader);
+    },
```
SandPod, line 118: Suggestion: Please make all input parameters named to explain on the call site what is being used for what.

test/headers/header_test.dart
```diff
@@ -1,152 +1,458 @@
...
+      singleTest('then valueOrNullIfInvalid is null',
+          header.valueOrNullIfInvalid, isNull);
+      singleTest('then value throws', () => header.value, throwsInvalidHeader);
+    },
+    variants: Headers.all,
+  );
+
+  const invalid = ['invalid'];
+  parameterizedGroup(
+    (v) => 'Given a "${v.key}" header with a raw value $invalid',
```
SandPod, line 144: Question: What does the `invalid` value symbolize and why is it important for the tests?

If it should represent an invalid value for the header, we could name the test `Given a "header" header with the invalid raw value "invalid"`

test/headers/header_test.dart
```diff
@@ -1,152 +1,458 @@
...
+      Headers.xPoweredBy,
+    }),
+  );
+
+  test(
+      'Given a headers object '
+      'when using the named extensions getters '
+      'then it succeeds', () {
+    final headers = Headers.empty();
+    // contentType, // Huh?
```
SandPod, line 188: Question: What is this comment for? Please remove or clarify.

test/headers/header_test.dart
```diff
@@ -1,152 +1,458 @@
...
+      Headers.trailer,
+      Headers.upgrade,
+      Headers.userAgent,
+      Headers.vary,
+      Headers.via,
+      Headers.xPoweredBy,
+    }),
+  );
+
+  test(
```
SandPod, line 183: Suggestion: If we break multiple scenarios in this test, then we will only get one repored at a time. This is either different tests or we should validate them all together (if the individual scenario is not important)

This also feels like the type of test that is not going to be updated IF we ever add more headers. Could we mitigate that in some way?

test/headers/header_test.dart
```diff
@@ -1,152 +1,458 @@
...
+    expect(headers.retryAfter, isNull);
+    expect(headers.server, isNull);
+    expect(headers.setCookie, isNull);
+    expect(headers.strictTransportSecurity, isNull);
+    expect(headers.vary, isNull);
+    expect(headers.wwwAuthenticate, isNull);
+    expect(headers.xPoweredBy, isNull);
+  });
+
+  test(
```
SandPod, line 252: Suggestion: Same comment here as above, please make it an explicit loop (where the individual value is ignored) or break it apart.

test/headers/header_test.dart
```diff
@@ -1,152 +1,458 @@
...
+      'Given a MutableHeader object '
+      'when using the named getter and setter extensions '
+      'then it succeeds', () {
+    Headers.build((mh) {
+      // contentType, // Huh?
```
SandPod, line 257: Suggestion: Please explain or remove.

test/util/test_util.dart
```diff
@@ -38,4 +37,40 @@ final _request = Request(RequestMethod.get, localhostUri);
...
+@isTest
+void parameterizedTest<T>(
+  String Function(T) descriptionBuilder,
+  void Function(T) body, {
+  required Iterable<T> variants,
+}) {
+  for (var v in variants) {
+    test(descriptionBuilder(v), () => body(v));
+  }
+}
```
SandPod, line 64: Suggestion: This is not used, please remove.

SandPod: Or use this for the "null" tests?

SandPod (commented):
A couple of additional comments.

It seems like the comments in the middle might not have been addressed.

test/headers/header_test.dart
nielsenko: We don't have a `contentType` header currently. The information is also available on the body, but I would expect it to possible to retrieve from the headers (/as well). What is your take on that?

BTW, the ContentType class is part of dart:io if I recall correctly, hence why I deferred it to the dart-io-decoupling pr. I should have put a TODO here.

test/headers/header_test.dart
nielsenko: See above

lib/src/headers/standard_headers_extensions.dart
nielsenko: > Question(For my knowledge): Shouldn't these collide with the static const variable in `headers.dart`?

In Dart you cannot access static members via an instance variable, so no `header.date` of type `DateTime?` is a extension property that uses the static member `Header.date` of type `HeaderFlyweight<DateTime>`.

> The comment also mentioned that these would not be part of headers.dart, what you mean by that?

I'm don't follow. Could you elaborate?

test/headers/header_test.dart
nielsenko: It will show up in the code-coverage report, if we miss one over time. I'm not sure how else to ensure that code is covered by tests.

test/headers/header_test.dart
nielsenko: Same as above

test/util/test_util.dart
nielsenko: It used for "null" tests (getter tests) now.

test/headers/header_test.dart
nielsenko: I tend to disagree. The arguments are not named on `expect` either. It is easy to lookup up what `singleTest` does since it is a super short function. Making the arguments named will make the call sites much more verbose. Can I persuade you otherwise on this?

test/headers/headers_flyweight_test.dart
nielsenko: These tests has been refactored extensive and given-when-then pattere applied

test/headers/headers_flyweight_test.dart
nielsenko: These tests has been refactored extensive and given-when-then pattern applied

lib/src/headers/headers.dart
```diff
@@ -1,14 +1,495 @@
...
+
+  Headers._(super.backing) : super._();
+
+  Headers transform(void Function(MutableHeaders) update) {
+    final mutable = MutableHeaders._from(this);
+    update(mutable);
+    return mutable._freeze();
+  }
+
+  // TODO: Should die
```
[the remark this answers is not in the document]
nielsenko: WIll happen in the dart-io-decoupling pr

lib/src/headers/header_flyweight.dart
nielsenko: This is mostly (inherited) defensive code. Those lines are not hit with the current implementation. Could consider removing them completely.

test/headers/header_test.dart
SandPod: Yeah, that part is a bit funky. I believe it is part of the body because wanted to automated deriving this from the body.

But I'm not sure that is the best approach. It makes sense that "all" headers can be set through the header class.

I'm open for nice suggestions to solve this.

test/headers/header_test.dart
SandPod: Aah, I forgot we have `CodeCov` now. Then this is probably fine 👍

test/headers/header_test.dart
SandPod: (We can solve this separately, though...)

SandPod (commented):
Two minor things and a couple of replies.

lib/src/headers/standard_headers_extensions.dart
nielsenko: You are right. Inside a class you can refer to static members without prefixing with class name, and hence it would be a conflict.

Do you feel it is to magic? Or is it okay as is?

test/headers/header_test.dart
nielsenko: I don't buy that hovering over the arguments on `singleTest` to see there meaning is in any way a problem for a contributor. 

However, I do feel that littering the call-sites with repeated use of named parameters will form a wall of text that gives little value, and makes it harder to see the important parts of the code.

Like `expect` `singleTest` is likely to be used repeatedly. Often you might even transition between the two forms during development. I prefer them to be as similar as possible.

Anyway, I have added a commit that does introduce named parameters to ground this discussion in something concrete. Do you still feel that is better?

lib/src/headers/standard_headers_extensions.dart
SandPod: No, this is OK! Just a knowledge gap for me! 👍

SandPod (approved):
LGTM! 🚀

Given that the "invalid" and named arguments for SingleTest are removed :)

test/headers/header_test.dart
nielsenko: Sorry, I read that to fast. Have updated description.

nielsenko:
@coderabbitai full review

nielsenko:
@coderabbitai review

nielsenko:
@coderabbitai review

lib/src/headers/exception/header_exception.dart
```diff
@@ -0,0 +1,72 @@
...
+  /// where it is expected or mandatory for proper processing.
+  const MissingHeaderException(super.description, {required super.headerType});
+
+  @override
+  String get httpResponseBody => 'Missing \'$headerType\' header';
+
+  @override
+  String toString() =>
+      'MissingHeaderException(description: $description, headerType: $headerType)';
+}
```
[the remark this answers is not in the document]
nielsenko: Will defer increasing code coverage for later

PR #42, opened 2025-04-03
fix: Ensure cache is updated immediately

nielsenko:
## Description

Previously reading a value through a header accessor inside an update callback would in some cases be stale. 
Example:
```dart
    final headers = Headers.build((mh) {
      mh.anInt = 42;
      expect(mh.anInt, 42); // first read goes well, since cache hangs of mh, but was empty for anInt
      mh.anInt = 1202;
      expect(mh.anInt, 1202); // <-- fail: second read would fail, cache still 42
    });
    // here cache is correct
    expect(headers.anInt, 1202);
```
- test: Add failing tests
- fix: Use raw-value as expando key, and update cache on set value

## Related Issues

Fix: #45 
 
## Pre-Launch Checklist

- [ ] I have updated/added relevant documentation (doc comments with `///`), ensuring consistency with existing project documentation.

## Breaking Changes
- [x] No breaking changes.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced header processing with improved caching strategies for better performance and reliability.
  
- **Tests**
  - Added test cases to ensure the caching logic works as intended, reducing redundant processing when header data remains unchanged.

### DIFF
--- a/lib/src/headers/header_accessor.dart
+++ b/lib/src/headers/header_accessor.dart
@@ -28,7 +28,8 @@ final class HeaderAccessor<T extends Object> {
   /// avoiding repeated parsing of the same raw header value.
   ///
   /// This is used internally by [getValueFrom] to implement the caching mechanism.
-  Expando get _cache => _caches.putIfAbsent(this, Expando.new);
+  Expando<T> get _cache =>
+      _caches.putIfAbsent(this, Expando<T>.new) as Expando<T>;
 
   /// The [key] is the name of the HTTP header.
   final String key;
@@ -66,14 +67,14 @@ final class HeaderAccessor<T extends Object> {
     T? Function(Exception)? orElse,
   }) {
     try {
-      var result = _cache[external] as T?;
-      if (result != null) return result; // found in cache
-
       final raw = external[key];
       if (raw == null) return null; // nothing to decode
 
+      var result = _cache[raw];
+      if (result != null) return result; // found in cache
+
       result = decode(raw);
-      _cache[external] = result;
+      _cache[raw] = result;
       return result;
     } on Exception catch (e) {
       if (orElse == null) _throwException(e, key: key);
@@ -93,8 +94,10 @@ final class HeaderAccessor<T extends Object> {
       Header((accessor: this, headers: external));
 
   /// [null] implies removing the header
-  void setValueOn(MutableHeaders external, T? value) =>
-      _setValue(external, key, value);
+  void setValueOn(MutableHeaders external, T? value) {
+    final raw = _setValue(external, key, value);
+    if (raw != null) _cache[raw] = value; // prime cache immediately
+  }
 
   /// Removes the header from the given [external] headers.
   void removeFrom(MutableHeaders external) => external.remove(key);
@@ -186,16 +189,16 @@ Never _throwException(
   );
 }
 
-void _setValue<T>(MutableHeaders headers, String key, T value) {
+Iterable<String>? _setValue<T>(MutableHeaders headers, String key, T value) {
   if (value == null) {
     headers.remove(key);
-  } else {
-    headers[key] = switch (value) {
-      String s => [s],
-      DateTime d => [formatHttpDate(d)],
-      TypedHeader t => [t.toHeaderString()],
-      Iterable<String> i => i,
-      Object o => [o.toString()],
-    };
+    return null;
   }
+  return headers[key] = switch (value) {
+    String s => [s],
+    DateTime d => [formatHttpDate(d)],
+    TypedHeader t => [t.toHeaderString()],
+    Iterable<String> i => i,
+    Object o => [o.toString()],
+  };
 }
--- a/test/headers/headers_accessor_test.dart
+++ b/test/headers/headers_accessor_test.dart
@@ -7,19 +7,13 @@ import 'headers_test_utils.dart';
 
 const _anInt = HeaderAccessor<int>(
   'anInt',
-  HeaderDecoderSingle(intParseAndCount),
+  HeaderDecoderSingle(parseInt),
 );
 
 const _someStrings = HeaderAccessor<List<String>>(
   'someStrings',
   HeaderDecoderMulti(parseStringList),
 );
-
-var _count = 0;
-int intParseAndCount(String s) {
-  ++_count;
-  return int.parse(s);
-}
 
 extension on Headers {
   int? get anInt => _anInt[this]();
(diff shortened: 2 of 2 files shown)


nielsenko:
@coderabbitai review

test/headers/headers_accessor_test.dart
```diff
@@ -145,4 +152,59 @@ void main() {
...
+      'Given a header accessor '
+      'when updating a value '
+      'then you can read the value immediately', () {
+    Headers.build((mh) {
+      mh.anInt = 42;
+      expect(mh.anInt, 42);
+      mh.anInt = 1202;
+      expect(mh.anInt, 1202);
+      mh['anInt'] = ['51']; // also for raw value updates
+      expect(mh.anInt, 51);
```
SandPod, line 166: Suggestion(NIT): Break this out into a separate test repeating the same scenario with a matching description that differentiates the tests.

test/headers/headers_accessor_test.dart
```diff
@@ -7,20 +7,27 @@ import 'headers_test_utils.dart';
...
-  HeaderDecoderSingle(int.parse),
+  HeaderDecoderSingle(intParseAndCount),
 );
 
 const _someStrings = HeaderAccessor<List<String>>(
   'someStrings',
   HeaderDecoderMulti(parseStringList),
 );
 
+var _count = 0;
```
SandPod, line 18: Suggestion: Make this count local inside of a single test.

Because this is a shared variable, we can't add another test without taking into account previous test runs. Please isolate this value so that new tests can rely on this starting out on 0.

SandPod: (This might mean that `_anInt` needs to be instantiated for each test.)

test/headers/headers_accessor_test.dart
```diff
@@ -145,4 +152,59 @@ void main() {
...
+  test(
+      'Given a header accessor '
+      'when reading the value '
+      'then decode is only called if raw value changed', () {
+    final headers = Headers.fromMap({
+      'anInt': ['102']
+    });
+
+    final offset = _count;
+    int decodeCount() => _count - offset; // compensate for previous calls
```
SandPod, line 179: Suggestion: This should not be required in the test logic.

test/headers/headers_accessor_test.dart
```diff
@@ -145,4 +152,59 @@ void main() {
...
+
+    expect(headers.anInt, 102); // read first time
+    expect(decodeCount(), 1); // one call to parse
+
+    expect(headers.anInt, 102); // read again
+    expect(decodeCount(), 1); // still only one call to parse, due to cache
```
SandPod, line 185: Suggestion: Break this test apart into a separate test. (Marking only a subset of lines but this applies to the whole test).

Please isolate the scenarios here so that we don't have this longer widely scoped test. Ideally, there should be no need for comments when a good description is supplied.

This specific test here is basically reading the value twice should result in a single call to decode count. The expect on the `read` value might not even be needed since that is not what is under test? We validate that in the test above right?

lib/src/headers/header_accessor.dart
```diff
@@ -93,8 +94,10 @@ final class HeaderAccessor<T extends Object> {
       Header((accessor: this, headers: external));
 
   /// [null] implies removing the header
-  void setValueOn(MutableHeaders external, T? value) =>
-      _setValue(external, key, value);
+  void setValueOn(MutableHeaders external, T? value) {
+    final raw = _setValue(external, key, value);
+    if (raw != null) _cache[raw] = value; // prime cache immediately
```
SandPod, line 99: Question: Shouldn't we "prime cache" also when we set the value to `null` (Meaning we remove the value right?)

lib/src/headers/header_accessor.dart
```diff
@@ -186,16 +189,16 @@ Never _throwException(
...
-      Object o => [o.toString()],
-    };
+    return null;
   }
+  return headers[key] = switch (value) {
+    String s => [s],
+    DateTime d => [formatHttpDate(d)],
+    TypedHeader t => [t.toHeaderString()],
+    Iterable<String> i => i,
+    Object o => [o.toString()],
```
SandPod, line 202: Question(For my knowledge): Are we limited to only supporting these value types? Any concern about something we do not cover here?

SandPod (commented):
Nice fix! A couple of questions to understand the implementation and as always, comments around test structure 🔧 ⚒️

test/headers/headers_accessor_test.dart
nielsenko: This test is for `Headers` the above is for `MutableHeaders`. Also, unlike the above test, it tests that multiple `Headers` objects, despite sharing state, still get the cached values right with a minimal number of decode calls.

This is not a good candidate for splitting as each expect depends on what happened before.

test/headers/headers_accessor_test.dart
nielsenko: Yes, if I want to use a const constructed `HeaderAccessor`. Note that internal to a test files, tests are never run in parallel, so this is stable.

lib/src/headers/header_accessor.dart
nielsenko: Yes, currently custom types will have to implement `TypedHeader` or do _the right thing_ in `.toString()`. This was what I was referring to when replacing `HeaderDecoder` with a `HeaderCodec` that also holds a _encoder_.

test/headers/headers_accessor_test.dart
nielsenko: Yes, you can add as many tests as you want, both before and after. This test will not be affected as it will always have a given `offset` on entry that we compensate for. Tests are never run in parallel inside a test file.

nielsenko: I have refactored the tests to construct a new `HeaderAccessor` per test, and track count in a local variable.

test/headers/headers_accessor_test.dart
nielsenko: I have refactored the tests to construct a new HeaderAccessor per test, and track count in a local variable.

test/headers/headers_accessor_test.dart
nielsenko: I have split this test-case into 3

lib/src/headers/header_accessor.dart
nielsenko: see #44 and #43

test/headers/headers_accessor_test.dart
```diff
@@ -145,4 +146,77 @@ void main() {
...
+      mh.anInt = 42;
+      expect(mh.anInt, 42);
+      mh.anInt = 1202;
+      expect(mh.anInt, 1202);
+      mh['anInt'] = ['51']; // also for raw value updates
+      expect(mh.anInt, 51);
+    });
+  });
+
+  group('Given a header accessor ', () {
```
SandPod, line 164: Suggestion(NIT): Change to `Given a header accessor that counts decodes`

SandPod (approved):
LGTM! 🚀

Last non blocking suggestion to better reflect the "Given" of the test.

PR #61, opened 2025-05-08
feat: Allow a trie to be attached as a subtrie to another

nielsenko:
## Description

Allows tries to be composed by attaching one below another.

## Related Issues
- Closes: #58

## Pre-Launch Checklist

## Breaking Changes
- [x] No breaking changes.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to extract sub-paths, get path length, and detect parameters in path segments.
  - Introduced a method to attach one route trie to another at a specified subpath, enabling more flexible route management.

- **Bug Fixes**
  - Improved error handling to prevent duplicate route registrations and invalid trie attachments.

- **Tests**
  - Added comprehensive tests for attaching tries, including parameterized routes and error scenarios.

### DIFF
--- a/test/router/path_trie_test.dart
+++ b/test/router/path_trie_test.dart
@@ -358,26 +358,75 @@ void main() {
       });
 
       test(
-          'Given a trie, '
+          'Given an empty trie, '
           'when attempting to attach another trie at the root path ("/"), '
-          'then throws ArgumentError', () {
+          'then it succeeds', () {
+        // This wasn't always so: https://github.com/serverpod/relic/pull/61/files#r2079670738
         expect(
           () => trieA.attach(NormalizedPath('/'), trieB),
-          throwsArgumentError,
-          reason: 'Cannot attach at root',
+          returnsNormally,
         );
       });
 
       test(
           'Given a trie with an existing path, '
-          'when attempting to attach another trie at that same path, '
-          'then throws ArgumentError', () {
-        trieA.add(NormalizedPath('/pathA/existing'), 1);
+          'when attempting to attach another trie on a sub-path, '
+          'then it succeeds and both tries are updated', () {
+        trieA.add(NormalizedPath('/pathA/existing/A'), 1);
+        trieB.add(NormalizedPath('/pathB/'), 2);
 
         expect(
           () => trieA.attach(NormalizedPath('/pathA/existing'), trieB),
+          returnsNormally,
+        );
+
+        // We can see B in A ..
+        var result = trieA.lookup(NormalizedPath('/pathA/existing/pathB'));
+        expect(result, isNotNull);
+        expect(result!.value, equals(2));
+
+        // .. but we can also see part of A in B
+        result = trieB.lookup(NormalizedPath('/A'));
+        expect(result, isNotNull);
+        expect(result!.value, equals(1));
+      });
+
+      test(
+          'Given a trie with an existing path, '
+          'when attempting to attach another trie at that same path that, '
+          'then it fails', () {
+        trieA.add(NormalizedPath('/existing'), 1);
+        trieB.add(NormalizedPath('/'), 2);
+        expect(
+          () => trieA.attach(NormalizedPath('/existing'), trieB),
           throwsArgumentError,
-          reason: 'Path already exists',
+          reason: 'Conflicting values',
+        );
+      });
+
+      test(
+          'Given a trie with an existing path, '
+          'when attempting to attach another trie at that same path that, '
+          'then it fails', () {
+        trieA.add(NormalizedPath('/existing/:a'), 1);
+        trieB.add(NormalizedPath('/:b'), 2);
+        expect(
+          () => trieA.attach(NormalizedPath('/existing'), trieB),
+          throwsArgumentError,
+          reason: 'Conflicting parameters',
+        );
+      });
+
+      test(
+          'Given a trie with an existing path, '
+          'when attempting to attach another trie at that same path that, '
+          'then it fails', () {
+        trieA.add(NormalizedPath('/existing/foo'), 1);
+        trieB.add(NormalizedPath('/foo'), 2);
+        expect(
+          () => trieA.attach(NormalizedPath('/existing'), trieB),
+          throwsArgumentError,
+          reason: 'Conflicting children',
         );
       });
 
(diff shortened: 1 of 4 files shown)


nielsenko:
@coderabbitai review

nielsenko:
@coderabbitai Fixed nits

test/router/path_trie_test.dart
```diff
@@ -330,5 +330,115 @@ void main() {
...
+        expect(result!.value, equals(10));
+        expect(result.parameters, isEmpty);
+      });
+
+      test(
+          'Given a trie, '
+          'when attempting to attach another trie at the root path ("/"), '
+          'then throws ArgumentError', () {
+        expect(
+          () => trieA.attach(NormalizedPath('/'), trieB),
```
SandPod, line 365: Question: Is the thinking that if this is desired you would simply use the other route? How about if we want to add multiple tries to a single root?

So two routes are exported from different packages with non-colliding routes  and I want to expand it all into the root?

test/router/path_trie_test.dart
```diff
@@ -330,5 +330,115 @@ void main() {
...
+      });
+
+      test(
+          'Given a trie with an existing path, '
+          'when attempting to attach another trie at that same path, '
+          'then throws ArgumentError', () {
+        trieA.add(NormalizedPath('/pathA/existing'), 1);
+
+        expect(
+          () => trieA.attach(NormalizedPath('/pathA/existing'), trieB),
```
SandPod, line 378: Question: What would happen if `trieB` would have the path `/pathA/existing` in this scenario?

I would expect this to result in us matching `/pathA/existing/pathA/existing`, especially in a "routing" scenario.

test/router/path_trie_test.dart
```diff
@@ -330,5 +330,115 @@ void main() {
...
+          'then parameters from trieB are correctly resolved', () {
+        // Setup trieB with a parameterized route
+        trieB.add(NormalizedPath('/items/:itemId'), 100);
+        trieA.attach(NormalizedPath('/api'), trieB); // Attach trieB at /api
+
+        // Lookup a path that goes through trieA into trieB
+        final result = trieA.lookup(NormalizedPath('/api/items/abc'));
+        expect(result, isNotNull);
+        expect(result!.value, equals(100));
+        expect(result.parameters, equals({#itemId: 'abc'}));
```
SandPod, line 413: Suggestion: add a test for when parameter collisions happen when attaching.

This might become a bit tricky to understand. Especially in a middleware context.

SandPod (commented):
Couple of comments and questions.

test/router/path_trie_test.dart
nielsenko: So to put in code..
```dart
final trieA = PathTrie<int>()..add('foo', 0);
final trieB = PathTrie<int>()..add('foo', 0);
final trieC = PathTrie<int>()..attach('/', trieA)..attach('/', trieB); // <-- you want this to succeed?
```

SandPod: I think `non-colliding` is the keyword here. Using your example:

```dart
final trieA = PathTrie<int>()..add('foo', 0);
final trieB = PathTrie<int>()..add('bar', 0);
final trieC = PathTrie<int>()..attach('/', trieA)..attach('/', trieB); // <-- This should succeed
```

nielsenko: How important is this? One of the reasons this PR is so relatively simple is that the `_root` node of the attached trie survives unscathed, and becomes a node in parent trie. This allows you to keep working on the combined trie via both parent and child.

test/router/path_trie_test.dart
nielsenko: Again, this keeps attach simple. But let me think about it once more.

nielsenko: But, if the path didn't already exist in `trieA` then yes, that would be the result.

test/router/path_trie_test.dart
nielsenko: Okay, I have implemented this. You can still keep working on both tries, as the attached trie's _root node is updated to match the updated node in the parent.

test/router/path_trie_test.dart
nielsenko: This will now work, as long as there are no conflicts wrt. value. parameter, and children names between the root node of `trieB` and the node at '/pathA/existing', which is not the case in this test.

test/router/path_trie_test.dart
```diff
@@ -358,26 +358,75 @@ void main() {
       });
 
       test(
-          'Given a trie, '
+          'Given an empty trie, '
           'when attempting to attach another trie at the root path ("/"), '
-          'then throws ArgumentError', () {
+          'then it succeeds', () {
+        // This wasn't always so: https://github.com/serverpod/relic/pull/61/files#r2079670738
```
SandPod, line 364: Suggestion: Since this is fixed at the introduction of this feature I think the comment that this wasn't always allowed can be removed.

We have not shipped a version where this was not allowed so it is not really important.

test/router/path_trie_test.dart
```diff
@@ -358,26 +358,75 @@ void main() {
...
+          () => trieA.attach(NormalizedPath('/existing'), trieB),
+          throwsArgumentError,
+          reason: 'Conflicting values',
+        );
+      });
+
+      test(
+          'Given a trie with an existing path, '
+          'when attempting to attach another trie at that same path that, '
+          'then it fails', () {
```
SandPod, line 410: Suggestion: Touch up description here.

SandPod: Same for the next test, to explain what they include.

SandPod (approved):
LGTM now!

Last minor thing to fix.

test/router/path_trie_test.dart
nielsenko: True. Just a small tribute to your astute eye. I will remove.

test/router/path_trie_test.dart
nielsenko: Okay 👍

test/router/path_trie_test.dart
nielsenko: Fixed

test/router/path_trie_test.dart
nielsenko: Fixed

PR #68, opened 2025-05-14
feat: Router middleware

nielsenko:
## Description
Introduce Routing Middleware and Enhance RequestContext

This PR introduces a new `RoutingMiddleware` to the Relic framework, enabling path-based request handling. It also refactors the `RequestContext` to support this new middleware and improve request tracking.

### Key Changes:

*   **New `RoutingMiddleware` (`relic/lib/src/middleware/routing_middleware.dart`):**
    *   Adds a `routeWith` function to easily integrate a `Router` into the request processing pipeline.
    *   The middleware inspects the request path and method, and if a route matches, it forwards the request to the corresponding handler.
    *   Path parameters extracted by the router are made available to handlers via an extension on `RequestContext` (`ctx.pathParameters`).

*   **`RequestContext` Enhancements (`relic/lib/src/adapter/context.dart`):**
    *   A `token` (an `Object`) has been added to `RequestContext`. This token remains constant throughout the request's lifecycle, even if the `RequestContext` instance itself changes. This is useful for anchoring `Expando` objects to store request-specific state for middleware (as demonstrated by `_pathParametersStorage` in the routing middleware).
    *   Constructors for `NewContext`, `ResponseContext`, and `HijackContext` now accept and store this `token`.
    *   The `request.toContext()` extension now requires a `token` argument.

*   **Example Update (`relic/example/example.dart`):**
    *   The main example has been updated to demonstrate the new routing capabilities.
    *   It now defines a route `/user/:name/age/:age` and a handler `hello` that uses path parameters.
    *   The pipeline now includes `logRequests()` and the new `routeWith(router)` middleware.
    *   A default "Not Found" handler is provided if no routes match.

*   **Server & Testing Updates:**
    *   **`relic/lib/src/relic_server.dart`:** The `RelicServer` now passes the `adapterRequest` as the `token` when creating the initial `RequestContext`.
    *   **`relic/test/static/test_util.dart` & `relic/test/util/test_util.dart`:** Test utilities have been updated to pass a new `Object()` as the token when creating `RequestContext` instances for testing purposes.

### Impact:

*   This change allows developers to define and manage routes within their Relic applications more effectively.
*   The introduction of the `token` in `RequestContext` provides a more robust way for middleware to associate state with a specific request.
*   The example application now showcases a more realistic use-case with path parameters and routing.

## Related Issues
- Closes: #55
- Closes: #57 

## Pre-Launch Checklist

- [ ] I have added new tests to verify the changes.

## Breaking Changes

The `Message` class (and hence `Request` and `Response`) no longer holds a context map. Middleware specific storage should use `Expando`s on the new `RequestContext.token`.

## Summary by CodeRabbit

- **New Features**
  - Added routing and middleware support, including request logging and path parameter extraction.
  - Introduced a new routing middleware for structured HTTP request handling and path parameter propagation.

- **Bug Fixes**
  - Improved 404 Not Found responses with custom messages for unmatched routes.

- **Refactor**
  - Enhanced request context handling to include a unique token for better request-specific state management.

- **Tests**
  - Updated test utilities to align with new context creation requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This package was born out of the needs of [Serverpod](https://serverpod.dev), as
 - We made everything type-safe (no more dynamic).
 - Encoding types have been moved to the `Body` of a `Request`/`Response` to simplify the logic when syncing up the headers and to have a single source of truth.
 - We've added parsers and validation for all commonly used HTTP headers. E.g., times are represented by `DateTime`, cookies have their own class with validation of formatting, etc.
-- Routing has been implemented using a `PathTrie` structure for efficient route matching and parameter extraction.
+- Routing has been implemented using a [trie](https://en.wikipedia.org/wiki/Trie) data-structure (`PathTrie`) for efficient route matching and parameter extraction.
 - Extended test coverage.
 - There are lots of smaller fixes here and there.
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ This package was born out of the needs of [Serverpod](https://serverpod.dev), as
 - We made everything type-safe (no more dynamic).
 - Encoding types have been moved to the `Body` of a `Request`/`Response` to simplify the logic when syncing up the headers and to have a single source of truth.
 - We've added parsers and validation for all commonly used HTTP headers. E.g., times are represented by `DateTime`, cookies have their own class with validation of formatting, etc.
+- Routing has been implemented using a `PathTrie` structure for efficient route matching and parameter extraction.
 - Extended test coverage.
 - There are lots of smaller fixes here and there.
 
@@ -25,7 +26,6 @@ Although the structure is very similar to Shelf, this is no longer backward comp
 
 Before a stable release, we're also planning on adding the following features:
 - We want to more tightly integrate a http server (i.e., start with `HttpServer` from `dart:io` as a base) with Relic so that everything uses the same types. This will also improve performance as fewer conversions will be needed.
-- Routing can be improved by using Radix trees. Currently, there is just a list being traversed, which can be an issue if you have many routes.
 - We're planning to add an improved testing framework.
 - Performance testing and optimizations.
 
@@ -33,33 +33,45 @@ In addition, we're planning to include Relic in [Serverpod](https://serverpod.de
 
 ## Example
 
-See `example/example.dart`
+See `relic/example/example.dart` for a runnable example. The following code demonstrates basic routing and request handling:
 
 ```dart
+import 'dart:io';
+
 import 'package:relic/relic.dart';
+import 'package:relic/src/adapter/context.dart';
+import 'package:relic/src/middleware/routing_middleware.dart';
 
-void main() async {
-  var handler =
-      const Pipeline().addMiddleware(logRequests()).addHandler(_echoRequest);
+/// A simple 'Hello World' server
+Future<void> main() async {
+  // Setup router
+  final router = Router<Handler>()..get('/user/:name/age/:age', hello);
 
-  var server = await serve(
-    handler,
-    RelicAddress.fromHostname('localhost'),
-    8080,
-  );
+  // Setup a handler.
+  //
+  // A [Handler] is function consuming and producing [RequestContext]s,
+  // but if you are mostly concerned with converting [Request]s to [Response]s
+  // (known as a [Responder] in relic parlor) you can use [respondWith] to
+  // wrap a [Responder] into a [Handler]
+  final handler = const Pipeline()
+      .addMiddleware(logRequests())
+      .addMiddleware(routeWith(router))
+      .addHandler(respondWith((final _) => Response.notFound(
+          body: Body.fromString("Sorry, that doesn't compute"))));
 
-  // Enable content compression
-  server.autoCompress = true;
+  // Start the server with the handler
+  await serve(handler, InternetAddress.anyIPv4, 8080);
 
-  print('Serving at http://${server.address.host}:${server.port}');
+  print('Serving at http://localhost:8080');
+  // Check the _example_ directory for other examples.
 }
 
-Response _echoRequest(Request request) {
-  return Response.ok(
-    body: Body.fromString(
-      'Request for "${request.url}"',
-    ),
-  );
+ResponseContext hello(final RequestContext ctx) {
+  final name = ctx.pathParameters[#name];
+  final age = int.parse(ctx.pathParameters[#age]!);
+
+  return (ctx as RespondableContext).withResponse(Response.ok(
+      body: Body.fromString('Hello $name! To think you are $age years old.')));
 }
 ```
 
--- a/example/example.dart
+++ b/example/example.dart
@@ -1,16 +1,25 @@
 import 'dart:io';
 
 import 'package:relic/relic.dart';
+import 'package:relic/src/adapter/context.dart';
+import 'package:relic/src/middleware/routing_middleware.dart';
 
 /// A simple 'Hello World' server
 Future<void> main() async {
+  // Setup router
+  final router = Router<Handler>()..get('/user/:name/age/:age', hello);
+
   // Setup a handler.
   //
   // A [Handler] is function consuming and producing [RequestContext]s,
   // but if you are mostly concerned with converting [Request]s to [Response]s
   // (known as a [Responder] in relic parlor) you can use [respondWith] to
   // wrap a [Responder] into a [Handler]
-  final handler = respondWith(hello);
+  final handler = const Pipeline()
+      .addMiddleware(logRequests())
+      .addMiddleware(routeWith(router))
+      .addHandler(respondWith((final _) => Response.notFound(
+          body: Body.fromString("Sorry, that doesn't compute"))));
 
   // Start the server with the handler
   await serve(handler, InternetAddress.anyIPv4, 8080);
@@ -18,6 +27,10 @@ Future<void> main() async {
   // Check the _example_ directory for other examples.
 }
 
-/// A very simple [Responder].
-Response hello(final Request request) =>
-    Response.ok(body: Body.fromString('Hello World'));
+ResponseContext hello(final RequestContext ctx) {
+  final name = ctx.pathParameters[#name];
+  final age = int.parse(ctx.pathParameters[#age]!);
+
+  return (ctx as RespondableContext).withResponse(Response.ok(
+      body: Body.fromString('Hello $name! To think you are $age years old.')));
+}
--- a/lib/src/middleware/routing_middleware.dart
+++ b/lib/src/middleware/routing_middleware.dart
@@ -1,0 +1,57 @@
+import '../adapter/context.dart';
+import '../handler/handler.dart';
+import '../method/request_method.dart';
+import '../router/router.dart';
+import 'middleware.dart';
+
+Middleware routeWith(final Router<Handler> router) =>
+    RoutingMiddleware(router).meddle;
+
+final _pathParametersStorage = Expando<Map<Symbol, String>>();
+
+class RoutingMiddleware {
+  final Router<Handler> _router;
+
+  RoutingMiddleware(this._router);
+
+  Handler meddle(final Handler next) {
+    return (final ctx) async {
+      final req = ctx.request;
+      final url = ctx.request.url; // TODO: Use requestUri
+      final match = _router.lookup(req.method.convert(), url.path);
+      if (match != null) {
+        ctx._pathParameters = match.parameters;
+        final handler = match.value;
+        return await handler(ctx);
+      } else {
+        return await next(ctx);
+      }
+    };
+  }
+}
+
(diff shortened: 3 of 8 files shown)


nielsenko:
@coderabbitai review

nielsenko:
@coderabbitai Add tests, please review

nielsenko:
@coderabbitai Updated README.md, please review

example/example.dart
```diff
@@ -1,23 +1,36 @@
...
   //
   // A [Handler] is function consuming and producing [RequestContext]s,
   // but if you are mostly concerned with converting [Request]s to [Response]s
   // (known as a [Responder] in relic parlor) you can use [respondWith] to
   // wrap a [Responder] into a [Handler]
-  final handler = respondWith(hello);
+  final handler = const Pipeline()
+      .addMiddleware(logRequests())
+      .addMiddleware(routeWith(router))
+      .addHandler(respondWith((final _) => Response.notFound(
```
tp, line 21: I was a bit surprised by the `addHandler` here. I would've assumed I could also write this as a middleware, since all the other routing ends up as a middleware.
So is this basically just a convenience function?

In the end I probably don't 100% get the disctiontion between `Handler` and `Middleware` (having read each types docs, but maybe there is some comparison somewhere else)?

Because the middleware could just as well modify (or add a new) response in the end, right? Though I would agree that it might be weird if the middleware would have to handle optional responses (if there was no handler), so a general `404` probably makes sense for a web server.

PS: Ah, I just wanted to see what happens if I remove the `addHandler` from the pipeline, but basically that is one required final call to convert it to the `Handler` type that `serve` needs.

lib/src/middleware/routing_middleware.dart
```diff
@@ -0,0 +1,57 @@
...
+    RoutingMiddleware(router).meddle;
+
+final _pathParametersStorage = Expando<Map<Symbol, String>>();
+
+class RoutingMiddleware {
+  final Router<Handler> _router;
+
+  RoutingMiddleware(this._router);
+
+  Handler meddle(final Handler next) {
```
tp, line 17: Should I know what `meddle` means in this context?

As a side-note, probably this project should also pull in lints like the Serverpod ones to document the public methods, right?

lib/src/relic_server.dart
```diff
@@ -93,7 +93,8 @@ class RelicServer {
     }
 
     try {
-      final ctx = request.toContext();
+      final ctx = request
+          .toContext(adapterRequest); // adaptor request will be the token
```
tp, line 97:
```suggestion
          .toContext(adapterRequest); // adapter request will be the token
```

Of just name the parameter if this is not clear here?

test/middleware/routing_middleware_test.dart
```diff
@@ -0,0 +1,423 @@
...
+          return (ctx as RespondableContext).withResponse(FakeResponse(404));
+        })(initialCtx);
+
+        expect(capturedParams, isNotNull);
+        expect(capturedParams, equals({#id: '123'}));
+        expect(resultingCtx, isA<ResponseContext>());
+        final response =
+            (resultingCtx as ResponseContext).response as FakeResponse;
+        expect(response.statusCode, equals(200));
+        expect(response.parameters, equals({#id: '123'}));
```
tp, line 73: Why thread the parameters through the response as well?

The `as FakeResponse` seems a bit weird, that the implementation object type has to be preserved even if it's not part of the interface.

I think asserting just on `capturedParams` is totally sufficient (and how I would write a similar test to see what the server was given).

test/middleware/routing_middleware_test.dart
```diff
@@ -0,0 +1,423 @@
...
+          return (ctx as RespondableContext)
+              .withResponse(FakeResponse(200, parameters: capturedParams));
+        }
+
+        router.add(Method.get, '/users', testHandler);
+
+        final initialCtx = FakeRequest('/users').toContext(Object());
+        final resultingCtx =
+            await middleware.meddle((final RequestContext ctx) async {
+          return (ctx as RespondableContext).withResponse(FakeResponse(404));
```
tp, line 92: This reads like it should override the response.


But I guess I am mistaken, and this is just another fallback handler?


I blame `meddle`, which could just mean anything from my point of view 😅🙈

test/middleware/routing_middleware_test.dart
```diff
@@ -0,0 +1,423 @@
...
+        bool handler1Called = false;
+        bool handler2Called = false;
+        bool fallbackCalled = false;
+
+        router1.add(Method.get, '/router1/:item',
+            (final RequestContext ctx) async {
+          handler1Called = true;
+          return (ctx as RespondableContext).withResponse(FakeResponse(201));
+        });
+        router2.add(Method.get, '/router2/:data',
```
tp, line 229: Is there any specificity matching (in and across routers), like `/x/:foo` vs. `/x/foo` (variable vs. literal?)

Maybe not matter what's currently implemented, do we have a test for the behavior, so that we would notice changes?

test/middleware/routing_middleware_test.dart
```diff
@@ -0,0 +1,423 @@
...
+        final initialCtx =
+            FakeRequest('/resource/abc/details/xyz').toContext(Object());
+        final resultingCtx = await pipeline(initialCtx);
+        final response =
+            (resultingCtx as ResponseContext).response as FakeResponse;
+
+        expect(nestedHandlerCalled, isTrue);
+        expect(response.statusCode, equals(200));
+        expect(capturedParams, isNotNull);
+        expect(capturedParams, equals({#resourceId: 'abc', #detailId: 'xyz'}));
```
tp, line 290: Interesting, given that the child route could not possibly know it will be mounted.

But at least the behavior is not specified and fixed.

tp (approved):
The actual code changes and tests in here look good.

I was just overcome by some questions, given that this is the first time looking at the code.

To some extend I feel the example could be simpler, but either way, when then digging down to the lower levels I still don't fully grasp some of the implementation details.

README.md
```diff
@@ -18,48 +18,60 @@ This package was born out of the needs of [Serverpod](https://serverpod.dev), as
 - We made everything type-safe (no more dynamic).
 - Encoding types have been moved to the `Body` of a `Request`/`Response` to simplify the logic when syncing up the headers and to have a single source of truth.
 - We've added parsers and validation for all commonly used HTTP headers. E.g., times are represented by `DateTime`, cookies have their own class with validation of formatting, etc.
+- Routing has been implemented using a `PathTrie` structure for efficient route matching and parameter extraction.
```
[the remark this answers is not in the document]
nielsenko: Maybe I should just have said trie. The `PathTrie` is called that way, since it is a trie over path segments. It also handles dynamic segments - not sure how to put that in the name. I'll find a good link. A place to start is: https://en.wikipedia.org/wiki/Trie.

Dear child has many names. _Tries_ are also often called Radix Trees.

I think it is quite common data-structure for routing. Both _routingkit_ and _spanner_ use some form of it as well.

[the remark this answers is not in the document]
nielsenko: I did consider doing a _Radix Tree_ (or _compressed prefix tree_) which is special kind of trie that compresses the path where possible. But I have a hunch that the added complexity doesn't really carry its weight for this use-case. I believe _spanner_ may do that, but we are still faster.

example/example.dart
```diff
@@ -1,23 +1,36 @@
...
+          body: Body.fromString("Sorry, that doesn't compute"))));
 
   // Start the server with the handler
   await serve(handler, InternetAddress.anyIPv4, 8080);
```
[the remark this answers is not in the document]
nielsenko: I'm not sure I follow. completely. 

Relic server will wrap you handler it a last ditch middleware, and turn most exceptions into 500 (or 403 if it is related to header parsing). We can elaborate on this last ditch middleware for sure.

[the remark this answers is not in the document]
nielsenko: And yes, you should wrap with top-level error handling middleware to control this behavior.

example/example.dart
```diff
@@ -1,23 +1,36 @@
...
+  final router = Router<Handler>()..get('/user/:name/age/:age', hello);
+
   // Setup a handler.
   //
   // A [Handler] is function consuming and producing [RequestContext]s,
   // but if you are mostly concerned with converting [Request]s to [Response]s
   // (known as a [Responder] in relic parlor) you can use [respondWith] to
   // wrap a [Responder] into a [Handler]
-  final handler = respondWith(hello);
+  final handler = const Pipeline()
```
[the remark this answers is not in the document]
nielsenko: I like this "short style". Basically the `with` wraps with a middleware as I understand.

lib/src/middleware/routing_middleware.dart
nielsenko: I have a `<X>Middleware` class that should produce a `Middleware` (`typedef Middleware = Handler Function(Handler)`). What should I call this function? I went with `meddle` since it is in the middle meddling with stuff. I'm open to suggestions 😄

lib/src/relic_server.dart
nielsenko: My english sits in the middle of the Atlantic ocean 😄

test/middleware/routing_middleware_test.dart
nielsenko: Yes, literal matching takes precedence. It is tested yes

example/example.dart
nielsenko: Yes. Think of handler like an atoms and middleware as composites.

Also the typedefs for the two function types are quite revealing:
```dart
typedef Handler = RequestContext Function(RequestContext);
typedef Middleware = Handler Function(Handler);
```

So middleware eats and produces handlers.

example/example.dart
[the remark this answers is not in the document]
nielsenko: @coderabbitai create an issue for this

test/middleware/routing_middleware_test.dart
nielsenko: I'll refactor to use `Pipeline` and `routeWith` for clarity.

example/example.dart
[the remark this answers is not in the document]
tp: Ups, I was mistaken 😅 The detailed error message is only logged to the server's console, not the response.

CleanShot 2025-05-16 at 10 37 23@2x


My bad, current approach seems good 👍

example/example.dart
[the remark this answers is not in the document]
tp: > I like this "short style". Basically the with wraps with a middleware as I understand.

Exactly. Concrete naming TBD, what I like about `Pipeline` I that the order is very clear (top to bottom -> outside in).

The `with` in the example above is a bit misleading in that `logRequest` would actually be the first thing to run. But maybe that is just the usual distinction between Handler and Middleware. And the whole builder set up should ensure that after a `with` not more handler is added (and vice versa), so at least on each Router-helper (like `app`/`server` above) there would only be one order in which things are registered.

test/middleware/routing_middleware_test.dart
nielsenko: Yes, this speaks to the difference between attaching routers as sub-routers of each other, and nesting middleware.
Since routers currently consume the full path, you have to attach them to nest, but it really should be possible to nest without attaching, and the nested router should only see the _remaining path_, and hence the parameters in the _remaining_ path.

This is something to revisit once #70 lands.

README.md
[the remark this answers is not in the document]
nielsenko: I'll update the readme.

test/middleware/routing_middleware_test.dart
nielsenko: Done

test/middleware/routing_middleware_test.dart
nielsenko: I no longer use the `_FakeResponse.parameters` in the test, and hence I have removed the class completely and just use the regular `Response` class.

README.md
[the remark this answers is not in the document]
nielsenko: Done

test/middleware/routing_middleware_test.dart
tp: Thanks!

test/middleware/routing_middleware_test.dart
tp: Nice, yeah, that simplifies comprehension in my eyes :)

example/example.dart
tp: Is there any other way to get from the `Pipeline` to a `Handler` to be used for `serve`?

CleanShot 2025-05-16 at 12 41 41@2x


That's why I was suggestion that final `build(withFallback: Handler)` call or similar in Slack.

nielsenko: Not today. This is inherited from shelf.

nielsenko: Will resolve for now. Thanks for all the great feedback.

PR #70, opened 2025-05-15
feat: PathTrie wildcard and tail matching support

nielsenko:
## Description

This change introduces support for wildcard (`*`) and tail (`**`) path segments in the `PathTrie` for more flexible route matching, along with related improvements to `NormalizedPath` and `Router`.

**Key Changes:**

*   **`PathTrie` Enhancements:**
    *   Introduces internal support for new dynamic segment types: `_Wildcard<T>` and `_Tail<T>`, alongside the existing `_Parameter<T>`.
    *   `LookupResult` now includes `matched` (the part of the path that matched up to and including dynamic segments) and `remaining` (the part of the path consumed by a `**` tail match).
    *   The `_build` method now handles the creation of nodes for `*` and `**` segments. It includes validation to ensure:
        *   `*` and `**` are standalone segments (e.g., `*foo` or `/**bar` are invalid).
        *   `**` (tail segment) must be the last segment in a path definition.
        *   Conflicting dynamic segment types cannot be defined at the same level in the trie.
    *   The `lookup` method is updated to:
        *   Recognize and match `*` and `**` segments.
        *   Correctly populate parameters, `matched` path, and `remaining` path in the `LookupResult`.
        *   Handle tail matches that consume zero remaining segments (e.g., route `/archive/**` matching lookup path `/archive`).

*   **`NormalizedPath` Improvements:**
    *   Added `@immutable` annotation.
    *   Introduced `NormalizedPath.empty` static instance for an empty path.
    *   Optimized `subPath` method to return `NormalizedPath.empty` for zero-length subpaths or `this` if the subpath covers the entire original path, improving efficiency by leveraging immutability.

*   **`Router` Update:**
    *   The `lookup` method in `Router` has been updated to correctly construct `LookupResult` with the new `matched` and `remaining` path fields from the `PathTrie`'s lookup result.

*   **Testing:**
    *   Adds new test files (`path_trie_tail_test.dart`, `path_trie_wildcard_test.dart`)

## Related Issues
- Closes: #72 
- Closes: #73 

## Benchmark results
2,4 GHz 8-Core Intel Core i9
```
./benchmark/benchmark.exe
Setting up benchmark data with 10000 routes...
Setup complete.
Starting benchmarks
---
Static Add
Routingkit Add Static x10000(RunTime): 10554.986013986014 us.
Spanner Add Static x10000(RunTime): 14277.356643356643 us.
Router Add Static x10000(RunTime): 7479.325373134328 us.
---
Static Lookup
Routingkit Lookup Static x10000(RunTime): 2231.669 us.
Spanner Lookup Static x10000(RunTime): 1741.7113943028485 us.
Router Lookup Static x10000(RunTime): 1523.3925 us.
---
Dynamic Add
Routingkit Add Dynamic x10000(RunTime): 14911.438461538462 us.
Spanner Add Dynamic x10000(RunTime): 59344.606060606064 us.
Router Add Dynamic x10000(RunTime): 10461.368644067798 us.
---
Dynamic Lookup
Routingkit Lookup Dynamic x10000(RunTime): 19026.396226415094 us.
Spanner Lookup Dynamic x10000(RunTime): 5818.2692307692305 us.
Router Lookup Dynamic x10000(RunTime): 3777.58 us.
Done
```

## Pre-Launch Checklist

## Breaking Changes
- [x] No breaking changes.

## Summary by CodeRabbit

- **New Features**
  - Enhanced routing to support wildcard (*) and tail wildcard (**) path segments for more flexible route matching.
  - Lookup results now provide detailed information about matched and remaining path segments.
  - Optimized path handling with immutable paths and reuse of empty path instances.

- **Bug Fixes**
  - Improved error handling and validation for invalid or conflicting route definitions.

- **Tests**
  - Added comprehensive tests for wildcard and tail wildcard matching, including precedence, parameter extraction, and error scenarios.

- **Chores**
  - Updated the meta package dependency for improved compatibility and stability.

### DIFF
--- a/lib/src/router/normalized_path.dart
+++ b/lib/src/router/normalized_path.dart
@@ -1,3 +1,5 @@
+import 'package:meta/meta.dart';
+
 import 'lru_cache.dart';
 
 /// Represents a URL path that has been normalized.
@@ -9,6 +11,7 @@ import 'lru_cache.dart';
 ///
 /// Instances are interned using an LRU cache for efficiency, meaning identical
 /// normalized paths will often share the same object instance.
+@immutable
 class NormalizedPath {
   /// Cache of interned instances
   static var interned = LruCache<String, NormalizedPath>(10000);
@@ -19,6 +22,9 @@ class NormalizedPath {
 
   /// Private constructor to create an instance with already normalized segments.
   NormalizedPath._(this.segments);
+
+  /// Empty normalized path instance.
+  static NormalizedPath empty = NormalizedPath._(const []);
 
   /// Creates a [NormalizedPath] from a given [path] string.
   ///
@@ -60,8 +66,14 @@ class NormalizedPath {
   ///
   /// The [start] parameter specifies the starting segment index (inclusive).
   /// The optional [end] parameter specifies the ending segment index (exclusive).
-  NormalizedPath subPath(final int start, [final int? end]) =>
-      NormalizedPath._(segments.sublist(start, end));
+  NormalizedPath subPath(final int start, [int? end]) {
+    end ??= length;
+    if (start == end) return NormalizedPath.empty;
+    if (start == 0 && end == length) {
+      return this; // since NormalizedPath is immutable
+    }
+    return NormalizedPath._(segments.sublist(start, end));
+  }
 
   /// The number of segments in this path
   int get length => segments.length;
--- a/lib/src/router/normalized_path.dart
+++ b/lib/src/router/normalized_path.dart
@@ -59,7 +59,7 @@ class NormalizedPath {
         result.add(segment);
       }
     }
-    return result;
+    return List.unmodifiable(result);
   }
 
   /// Returns a new [NormalizedPath] representing a subpath of this path.
--- a/test/router/path_trie_crud_test.dart
+++ b/test/router/path_trie_crud_test.dart
@@ -175,6 +175,61 @@ void main() {
         expect(() => trie.update(pathDefinition, 1), throwsArgumentError);
         expect(trie.lookup(NormalizedPath('/articles/any')), isNull);
       });
+
+      test(
+          'Given a trie with an existing wildcard path /data/* and value, '
+          'when update is called for /data/* with a new value, '
+          'then lookup for a matching path returns the new value', () {
+        final pathDefinition = NormalizedPath('/data/*');
+        trie.add(pathDefinition, 1);
+
+        trie.update(pathDefinition, 2);
+
+        // Verify by looking up a path that would match the wildcard definition
+        final result = trie.lookup(NormalizedPath('/data/something'));
+        expect(result, isNotNull);
+        expect(result!.value, equals(2));
+        expect(
+            result.parameters, isEmpty); // Wildcards don't produce parameters
+        expect(result.matched.toString(), '/data/something');
+        expect(result.remaining.segments, isEmpty);
+      });
+
+      test(
+          'Given a trie, '
+          'when update is called for a wildcard path /data/* that does not exist as a defined route, '
+          'then an ArgumentError is thrown', () {
+        final pathDefinition = NormalizedPath('/data/*');
+        expect(() => trie.update(pathDefinition, 1), throwsArgumentError);
+        expect(trie.lookup(NormalizedPath('/data/anything')), isNull);
+      });
+
+      test(
+          'Given a trie with an existing tail path /files/** and value, '
+          'when update is called for /files/** with a new value, '
+          'then lookup for a matching path returns the new value', () {
+        final pathDefinition = NormalizedPath('/files/**');
+        trie.add(pathDefinition, 1);
+
+        trie.update(pathDefinition, 2);
+
+        // Verify by looking up a path that would match the tail definition
+        final result = trie.lookup(NormalizedPath('/files/a/b.txt'));
+        expect(result, isNotNull);
+        expect(result!.value, equals(2));
+        expect(result.parameters, isEmpty);
+        expect(result.matched.toString(), '/files');
+        expect(result.remaining.toString(), '/a/b.txt');
+      });
+
+      test(
+          'Given a trie, '
+          'when update is called for a tail path /files/** that does not exist as a defined route, '
+          'then an ArgumentError is thrown', () {
+        final pathDefinition = NormalizedPath('/files/**');
+        expect(() => trie.update(pathDefinition, 1), throwsArgumentError);
+        expect(trie.lookup(NormalizedPath('/files/anything/else')), isNull);
+      });
     });
 
     group('remove', () {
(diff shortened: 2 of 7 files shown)


nielsenko:
@coderabbitai review

nielsenko:
@coderabbitai review

lib/src/router/normalized_path.dart
```diff
@@ -9,6 +11,7 @@ import 'lru_cache.dart';
 ///
 /// Instances are interned using an LRU cache for efficiency, meaning identical
 /// normalized paths will often share the same object instance.
+@immutable
 class NormalizedPath {
   /// Cache of interned instances
   static var interned = LruCache<String, NormalizedPath>(10000);
```
tp, line 17: Q: Will this class also be used for incoming requests (as I didn't find any current usage outside of the router)?

Just wondering where the "duplicate" instances would come from that we need _internal_ caching.

lib/src/router/normalized_path.dart
```diff
@@ -60,8 +66,14 @@ class NormalizedPath {
   ///
   /// The [start] parameter specifies the starting segment index (inclusive).
   /// The optional [end] parameter specifies the ending segment index (exclusive).
-  NormalizedPath subPath(final int start, [final int? end]) =>
-      NormalizedPath._(segments.sublist(start, end));
+  NormalizedPath subPath(final int start, [int? end]) {
+    end ??= length;
+    if (start == end) return NormalizedPath.empty;
+    if (start == 0 && end == length) {
+      return this; // since NormalizedPath is immutable
```
tp, line 73: Suggestion: Maybe use `List.unmodifiable(result)` in `_normalize`?

Just to be safe for future changes and external behavior. Right now everything _inside_ this calls seems okay (as we always create our own `List` from the `path` `String`). 

But `segments` is still publicly exposed…

test/router/path_trie_crud_test.dart
```diff
@@ -175,6 +175,61 @@ void main() {
...
       });
+
+      test(
+          'Given a trie with an existing wildcard path /data/* and value, '
+          'when update is called for /data/* with a new value, '
+          'then lookup for a matching path returns the new value', () {
+        final pathDefinition = NormalizedPath('/data/*');
+        trie.add(pathDefinition, 1);
+
+        trie.update(pathDefinition, 2);
```
tp, line 186: Q: We does this discern between `add` and `update` compared to a just a "set" on a `Map`?

test/router/path_trie_crud_test.dart
```diff
@@ -300,6 +355,72 @@ void main() {
...
+        final removedValue = trie.remove(pathDefinition);
+
+        expect(removedValue, equals(30));
+        expect(trie.lookup(NormalizedPath('/files/a/b/c')), isNull,
+            reason: 'Tail path should be removed.');
+        final specificLookup =
+            trie.lookup(NormalizedPath('/files/specific/file.txt'));
+        expect(specificLookup?.value, 40,
+            reason:
+                'More specific path should ideally remain if structured correctly.');
```
tp, line 410:
```suggestion
                'More specific path should remain');
```

Or what does "structured correctly" mean? Is it not just "the code working correctly" / according to spec?

test/router/path_trie_tail_test.dart
```diff
@@ -0,0 +1,265 @@
...
+        'Given a trie with path /static/**, '
+        'when /static/css/style.css is looked up, '
+        'then it matches with correct value and remaining path', () {
+      trie.add(NormalizedPath('/static/**'), 1);
+      final result = trie.lookup(NormalizedPath('/static/css/style.css'));
+      expect(result, isNotNull);
+      expect(result!.value, 1);
+      expect(result.parameters, isEmpty);
+      expect(result.matched.path, '/static');
+      expect(result.remaining.path, '/css/style.css');
```
tp, line 23: Where does the requirement come from that `path` above (and in old code) and `remaining` start with `/`?

I would think in both cases they are relative paths, as the "route" (in practice) does not know what it's parent / mount point will be, and the remaining path / sub segment is definitely relative to it's parent.

I think the static file serving is a good example.
Does Dart have a trasversal safe API where you could do something like `Directory('/data/static').resolveSafe(userInputFilePath)`? (Go added this recently.)

Then I would maybe try to see to give a value that matches this and can be nicely forwarded.

Or maybe my hunch is wrong here, and even "relative" paths have a benefit of starting with `/`.

test/router/path_trie_tail_test.dart
```diff
@@ -0,0 +1,265 @@
...
+    });
+
+    test(
```
tp, line 26: Nit: This is basically the same as above, just with more "sub folders", right?

At first I was looking for another different as the switch from `static` to `files` etc., but I don't think there is one.

test/router/path_trie_tail_test.dart
```diff
@@ -0,0 +1,265 @@
...
+      expect(result.matched.path, '/files');
+      expect(result.remaining.path, '/a/b/c/doc.txt');
+    });
+
+    test(
+        'Given a trie with path /archive/**, '
+        'when /archive/ (ends where tail starts) is looked up, '
+        'then it matches with an empty remaining path', () {
+      trie.add(NormalizedPath('/archive/**'), 1);
+      final result = trie.lookup(NormalizedPath('/archive/'));
```
tp, line 44: ~~What if this were just `/archive` (no trailing slash)?~~

Again confused that below is basically testing exactly this, but using another naming again.

Personally I would probably do better with parameterized test tables, but that might not be the convention for this project.

test/router/path_trie_tail_test.dart
```diff
@@ -0,0 +1,265 @@
...
+        trie.add(NormalizedPath('/test/**'), 1);
+        expect(
+            () => trie.add(NormalizedPath('/test/*'), 2), throwsArgumentError);
+      });
+
+      test(
+          'Given a trie with /test/*, '
+          'when adding /test/** (tail after wildcard at same level), '
+          'then an ArgumentError is thrown', () {
+        trie.add(NormalizedPath('/test/*'), 1);
```
tp, line 249: Q: Was this "your choice" or an implementation detail?

I suppose one could still argue that `/*` is more precise for `/foo` than `/**`, but either way, I don't see this relevant in practice. Just wondered whether that would make the internal implementation more complicated (for little gain), or is just a safeguard against confusing usage.

test/router/path_trie_wildcard_test.dart
```diff
@@ -0,0 +1,154 @@
...
+      final result = trie.lookup(NormalizedPath('/a/b/c/d'));
+      expect(result, isNotNull);
+      expect(result!.value, 1);
+      expect(result.parameters, isEmpty);
+      expect(result.matched.path, '/a/b/c/d');
+      expect(result.remaining.segments, isEmpty);
+    });
+
+    test(
+        'Given a trie with path /a/*/b, '
```
tp, line 66: Is there a test for `/a/*` being requested with just `/a(/)` (should return null)? I don't see one.

PS: I assume trailing slashes are always optional? Or we just assume that these don't change the behavior (so that we don't have to test that variant. At least `//` is collapsed always per the normalization).

tp (approved):
👏 Nice work!

Just a few questions for my understanding really 😀

lib/src/router/normalized_path.dart
nielsenko: .. if it doesn't hurt the benchmark too much 😃

test/router/path_trie_crud_test.dart
nielsenko: Yes, there are `add`, `update`, and `addOrUpdate` with the expected semantics. The first two will throw, if a value is already attached (`add`) , or missing (`update`).

test/router/path_trie_tail_test.dart
nielsenko: Maybe I'm wrong, but don't the path component of an url always start with `/`?

test/router/path_trie_tail_test.dart
nielsenko: Actually I do have a `parameterizedTest`  function in this project. I should use that here.

test/router/path_trie_wildcard_test.dart
nielsenko: There is:
```dart
   test(
        'Given a trie with path /a/*/b, '
        'when /a/b (fewer segments) is looked up, '
        'then no match is found', () {
      trie.add(NormalizedPath('/a/*/b'), 1);
      expect(trie.lookup(NormalizedPath('/a/b')), isNull);
    });
```
if that is what you mean.

And yes, trailing slashes will be removed by normalization.

test/router/path_trie_tail_test.dart
nielsenko: Not sure I follow. This is the very first test in the file.

lib/src/router/normalized_path.dart
tp: >  It is quite common for a server to route the same paths repeatedly.

Yeah, I was wondering whether I missed that (that the normalized path is also used in incoming requests), and not just when building the router tree.

test/router/path_trie_tail_test.dart
tp: It is the second one, and basically (I think) just a variant / enhancement of the first, but since it was using different names for the use-case under test (`/files` instead of `/static`) I at first thought they were more different than they turned out to be.

But no big deal, just that I think that having the same setup could help seeing that's actually under test (and thus always using the same "example", or just `/a/b/c.d` consistently).

test/router/path_trie_wildcard_test.dart
tp: Well, it's a little bit different, in now the route does not match because the suffic `/b` is missing.

I was looking for a test case like

```
   trie.add(NormalizedPath('/a/*'), 1);
   expect(trie.lookup(NormalizedPath('/a')), isNull);
```

test/router/path_trie_tail_test.dart
tp: > Maybe I'm wrong, but don't the path component of an url always start with /?

I think there is an explicit distinction between absolute and relative URIs, and I just wondered whether by definition not all of these in there would be relative.

lib/src/router/normalized_path.dart
nielsenko: Not sure what happens in
```dart
external factory List.unmodifiable(Iterable elements);
```
but that was a lot less costly than I feared.

test/router/path_trie_tail_test.dart
nielsenko: To elaborate with an example

Consider:
1. /*/foo
2. /**

Lookup /bar/nofoo/

lib/src/router/normalized_path.dart
nielsenko: Done

test/router/path_trie_tail_test.dart
nielsenko: I have gone over the tests and removed "duplicates"

nielsenko:
Resolved all comments to merge

PR #74, opened 2025-05-16
docs: Design typed pipeline

nielsenko:
## Description
This PR introduces a design for a new typed pipeline system in Relic. The goal is to improve both developer ergonomics (clear, type-safe access to context data like `context.user`) and compile-time safety, preventing errors from misconfigured middleware. The core idea involves `ContextProperty`, extension type views, and a generics based `PipelineBuilder`. See [DESIGN_typed_context_pipeline.md](https://github.com/serverpod/relic/blob/design-typed-pipeline/design/DESIGN_typed_context_pipeline.md).

### DIFF
--- a/design/DESIGN_typed_context_pipeline.md
+++ b/design/DESIGN_typed_context_pipeline.md
@@ -310,10 +310,142 @@ This typed pipeline introduces a shift from the traditional `Middleware = Handle
     *   **Gained**: Strong compile-time type safety for the data flowing through the request context. This significantly reduces a class of runtime errors due to misconfigured pipelines.
     *   **Different Flexibility**: Some dynamic flexibility found in the `Handler Function(Handler)` pattern (e.g., complex around-logic, dynamically choosing the next handler in the chain from within a middleware) is handled differently (e.g., via exceptions, or by moving logic into handlers). For many common middleware tasks (logging, data enrichment, simple auth checks), the typed pipeline offers a clearer and safer model.
 
-## 8. General Considerations
+## 8. Integrating with Routing
+
+The typed context pipeline is designed to prepare a rich, type-safe context that can then be utilized by a routing system to dispatch requests to appropriate endpoint handlers. Relic's `Router<T>` class can be seamlessly integrated with this pipeline.
+
+The core idea is that the `PipelineBuilder` sets up a chain of common middleware (e.g., authentication, session management, logging). The final function passed to `PipelineBuilder.build(...)` will be a "routing dispatcher." This dispatcher uses the context prepared by the common middleware, performs route matching, potentially adds route-specific data (like path parameters) to the context, and then executes the endpoint handler chosen by the router.
+
+### 8.1. Context for Route Parameters
+
+Endpoint handlers often need access to path parameters extracted during routing (e.g., the `:id` in `/users/:id`). A `ContextProperty` and corresponding view should be defined for these.
+
+```dart
+// Example: Data class for route parameters
+class RouteParameters {
+  final Map<Symbol, String> params;
+  RouteParameters(this.params);
+
+  String? operator [](Symbol key) => params[key];
+  // Potentially add other useful accessors
+}
+
+// Example: Private ContextProperty for route parameters
+// (Typically defined in a routing-related module/library)
+final _routeParametersProperty =
+  ContextProperty<RouteParameters>('relic.routing.parameters');
+```
+
+### 8.2. Context Views for Endpoint Handlers
+
+Endpoint handlers will require a context view that provides access to both the common context data (prepared by the initial pipeline) and the specific `RouteParameters`.
+
+```dart
+// Example: A view that combines UserContext (from common pipeline) and RouteParameters
+// (Assumes UserContextView is already defined)
+extension type UserRouteContextView(RequestContext _relicContext) implements UserContextView {
+  RouteParameters get routeParams => _routeParametersProperty.get(_relicContext);
+
+  // This method will be called by the routing dispatcher after parameters are extracted.
+  void attachRouteParameters(RouteParameters params) {
+    _routeParametersProperty.set(_relicContext, params);
+  }
+}
+
+// Other combinations can be created as needed (e.g., BaseRouteContextView, UserSessionRouteContextView).
+```
+
+### 8.3. Router's Generic Type `T`
+
+The generic type `T` in `Router<T>` will represent the actual endpoint handler functions. These functions will expect an enriched context view that includes common context data and route parameters.
+
+For example, `T` could be:
+`FutureOr<Response> Function(UserRouteContextView context)`
+
+### 8.4. The Routing Dispatcher Function
+
+This function is passed to `PipelineBuilder.build(...)`. It receives the context prepared by the common middleware chain (e.g., `UserContextView`). Its responsibilities are:
+1.  Use the incoming request details (from the context) to perform a route lookup via `Router<T>`.
+2.  Handle cases where no route is matched (e.g., by throwing a `RouteNotFoundException` or returning a 404 `Response`).
+3.  If a route is matched, extract the endpoint handler and any path parameters.
+4.  Create the specific context view required by the endpoint handler (e.g., `UserRouteContextView`), attaching the extracted `RouteParameters` to it.
+5.  Execute the chosen endpoint handler with this enriched context.
+
+```dart
+// Example: Routing Dispatcher
+// Assume 'myAppRouter' is an instance of Router<FutureOr<Response> Function(UserRouteContextView)>
+// Assume 'UserContextView' is the output view from the common middleware pipeline.
+
+FutureOr<Response> routingDispatcher(UserContextView commonContext) {
+  final request = commonContext.request;
+
+  // Perform route lookup using Relic's Router.
+  final lookupResult = myAppRouter.lookup(
+      request.method.convert(), // Or however method is represented
+      request.uri.path);
+
+  if (lookupResult == null || lookupResult.value == null) {
+    // Option 1: Throw a specific RouteNotFoundException for centralized error handling.
+    throw RouteNotFoundException(
+        'Route not found for ${request.method.value} ${request.uri.path}');
+    // Option 2: Directly return a 404 Response (less flexible for global error handling).
+    // return Response.notFound(...);
+  }
+
+  final endpointHandler = lookupResult.value;
+  final pathParams = RouteParameters(lookupResult.parameters);
+
+  // Create the specific context view for the endpoint handler by wrapping the same
+  // underlying RequestContext and attaching the extracted route parameters.
+  final endpointContext = UserRouteContextView(commonContext._relicContext);
+  endpointContext.attachRouteParameters(pathParams);
+
+  // Execute the chosen endpoint handler.
+  return endpointHandler(endpointContext);
+}
+```
+
+### 8.5. Pipeline Setup with Routing
+
+The `PipelineBuilder` is used to construct the common middleware chain, with the `routingDispatcher` as the final step.
+
+```dart
+// Example: In your server setup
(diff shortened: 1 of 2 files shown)


nielsenko:
@coderabbitai could you review the design document in DESIGN_typed_context_pipeline.md

nielsenko:
@coderabbitai Added extra section on routing as well as a larger runnable sample in appendix_a.dart. Could update your review?

design/DESIGN_typed_context_pipeline.md
```diff
@@ -0,0 +1,451 @@
...
+
+The system relies on a few key components:
+
+### 2.1. Relic's `RequestContext` and Stable Request Token
+
+This design utilizes Relic's existing `RequestContext` (defined in `relic/lib/src/adapter/context.dart`). The key aspects of `RequestContext` relevant to this design are:
+*   It provides access to the current `Request` object (e.g., via a `request` getter).
+*   It contains a stable `token` (e.g., via a `token` getter). This `token` is a unique identifier for the entire lifecycle of a single request and remains constant even if the `RequestContext` instance itself undergoes state transitions (e.g., from `NewContext` to `ResponseContext`).
+
+All request-scoped data attached via `Expando` (managed by `ContextProperty`) will be keyed off this stable `requestToken`. The extension type views defined in this design will wrap an instance of Relic's `RequestContext`.
```
tp, line 23: Not sure about the performance characteristics with using the various expandos. I think we could similarly have just the state on the context / token itself, and then provide getter / setter like `V get(ContextProperty<V> prop)` on it.

I like how easy it is in the current approach to hide the setter though (just making that part file-local for example), and wonder if replicating this would be much harder in my above suggestion (e.g. you could have something like `Completer`, with an "value reader" and "value setter" being distinct, and only 1 exported as a direct helper).

design/DESIGN_typed_context_pipeline.md
```diff
@@ -0,0 +1,451 @@
...
+  Session get session => _sessionProperty.get(_relicContext);
+
+  void attachSession(Session session) {
+    _sessionProperty.set(_relicContext, session);
+  }
+}
+final _sessionProperty = ContextProperty<Session>('relic.session');
+
+// A composite view indicating both User and Session information are available.
+extension type UserSessionContextView(RequestContext _relicContext) implements UserContextView, SessionContextView {
```
tp, line 139: I think this is the Achilles' heel of the proposal.

While the application could easily build "the one full context" (per router), how would this combine with externally provided middle?

Wouldn't I have to wrap each step in a `*ContextView` that threads through all the previous one (specific to my current app configuration)?

design/DESIGN_typed_context_pipeline.md
```diff
@@ -0,0 +1,451 @@
...
+```
+
+## 6. Benefits
+
+*   **Compile-Time Safety**: The primary goal. Misconfigured pipelines (e.g., missing middleware, incorrect order affecting context data) are caught by the Dart analyzer.
+*   **Improved Developer Ergonomics**:
+    *   Handlers and middleware can declare precisely the context view (and thus data) they expect.
+    *   Access to context data via extension type getters is type-safe and clear (e.g., `context.user`).
+*   **Minimal Runtime Overhead for Views**: Extension types are intended to be zero-cost compile-time wrappers. The `ContextProperty` helper encapsulates `Expando` lookups/attachments, which are generally efficient.
+*   **No Class Explosion**: Avoids needing a distinct context `class` for every possible combination of middleware. Extension types provide views, and `ContextProperty` manages data association with the stable `requestToken`.
```
tp, line 290: > Avoids needing a distinct context `class` for every possible combination of middleware.

Is this true though? Maybe I misunderstand some aspect of the implementation or how I could avoid that, but I feel like I would always need to make the "next" middleware aware of the previous (and thus alter it's type when pre-pending a new one). (Assuming all middleware enhances the context with local data and is not just a passthrough like the logger in this example.)

design/DESIGN_typed_context_pipeline.md
```diff
@@ -0,0 +1,451 @@
...
+*   **Clarity and Documentation**: The type signatures of middleware and handlers explicitly document their context dependencies. View methods (e.g., `attachUser`) and `ContextProperty` provide clear, discoverable APIs for data management.
+*   **Modularity & Encapsulation**: `ContextProperty` encapsulates `Expando` usage. Modules define their data properties cleanly.
+
+## 7. Middleware Paradigm Shift and Implications
+
+This typed pipeline introduces a shift from the traditional `Middleware = Handler Function(Handler innerHandler)` pattern previously used. Understanding these changes is crucial:
+
+*   **New Middleware Signature**: Middleware in this design are functions with a signature like `OutputView Function(InputView)`. They transform context views rather than wrapping an inner handler.
```
tp, line 298: Getting here would be awesome 🤩

design/DESIGN_typed_context_pipeline.md
```diff
@@ -0,0 +1,451 @@
...
+*   **Linear Chain**: The `PipelineBuilder` composes middleware into a linear chain of context transformations. Each middleware is expected to process the context and pass control (via its return value) to the next stage defined by the builder.
+*   **Short-Circuiting (e.g., Denying Access)**:
+    *   Middleware should not directly return a `Response` to short-circuit the pipeline.
+    *   Instead, if a middleware needs to stop processing and return an error (e.g., an authentication middleware denying access due to an invalid token), it should **throw a specific exception** (e.g., `AuthorizationRequiredError("Invalid token")`, `PaymentRequiredError()`).
```
tp, line 302: Tricky, personally I was not a fan of the `throw`.

But maybe this could be handle with some `Result`-ish type which distinct between "early return" and "next".

But either way, I think that is a secondary consideration and would not need to happen in combination with the major change of this proposal, right?

design/DESIGN_typed_context_pipeline.md
```diff
@@ -0,0 +1,451 @@
...
+
+## 9. General Considerations
+
+*   **PipelineBuilder Complexity**: The implementation of `PipelineBuilder`, especially its generic typing, is somewhat complex, but this complexity is encapsulated for the end-user.
+*   **Boilerplate for `ContextProperty` and Views**: Each new piece of context data requires defining a `ContextProperty` instance and corresponding view methods. However, this is more structured and less error-prone than raw `Expando` usage.
+*   **Learning Curve**: Developers using the framework will need to understand context views, `ContextProperty`, the role of `requestToken`, the pipeline builder, and the implications of the new middleware paradigm.
+*   **Discipline with `requestToken`**: The `ContextProperty` helper ensures that data is keyed off the stable `token` within the `RequestContext`, mitigating direct misuse of `Expando`s with transient `RequestContext` instances themselves as keys.
+*   **Middleware Return Types**: Middleware authors must be careful to return the correct context view type that accurately reflects the data they've attached via `ContextProperty` and the `requestToken`.
```
tp, line 446: Maybe this already refers to the issue I have encountered below, where I could get the code to compile but the middleware implementation to be wrong and err at runtime.

I suppose "just" needing to take care in the middleware implementation (which should mostly be small) is still a win over having to manually match types across the entire router setup though.

So maybe this is the right trade-off. For comparison though, maybe we should take a look at what costs a fully type-safe middleware implementation would really entail. E.g. I imagine that even a simple middleware like an auth check or custom logger would access a network store (to validate the session), or do a big computation (to verify a token), such that creating a single class instance to store/wrap the result would not be a practical slow-down at all.

tp (commented):
I applaud what you're trying to achieve here 😀

It's really unfortunate that many of the "obvious" things we would want to do with generics here (like having an add that is `ExistingMiddleware & T addMiddleware(Middleware<T> middleware)`) is currently not possible.

I was not a big fan of the `mixin`s in each route, but I wonder if the higher-level like a `Router` / `Pipeline` is exactly where that would make sense and solve a ton of type-issues and where stuff would only have to be added/changed in a single place.
(Having applications subclass `Pipline with A, B, C` is probably the easiest one, another might be to have an explicit middleware step (method) that outputs a new context (by calling into the middleware). Offers more flexibility, but less out of the box "speed".)

Just my 2 cents, will have to ponder this for a bit.
Also many of the ideas that came to my mind for how to handle this, just don't work out nicely in practice…

Looking forward to having a chat about this :)

design/DESIGN_typed_context_pipeline.md
nielsenko: Yes, having just fought to get rid of `HijackException` based control flow I know where you are coming from.

However, you won't inadvertently catch these exceptions in the pipeline as the middleware don't wrap each other as today.

I kind of made peace with it. Passing a context to a handler that it shouldn't react to is also a bit weird, and then there is the question of typing.

design/DESIGN_typed_context_pipeline.md
nielsenko: Looking forward to discuss your thoughts on this.

design/appendix_a.dart
```diff
@@ -0,0 +1,246 @@
...
+  final Object token; // Stable unique token
+  RequestContext(this.request, this.token);
+}
+
+class NewContext extends RequestContext {
+  NewContext(super.request, super.token);
+}
+
+// === ContextProperty and Views Stubs ===
+class ContextProperty<T extends Object> {
```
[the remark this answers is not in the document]
nielsenko: Also, this pattern makes sense with the current paradigm as well.

design/appendix_a.dart
```diff
@@ -0,0 +1,246 @@
...
+
+// Admin data and view
+class AdminRole {
+  final String roleName;
+  AdminRole(this.roleName);
+}
+
+final _adminRoleProperty = ContextProperty<AdminRole>('admin_role');
+extension type AdminContextView(RequestContext _relicContext)
+    implements UserContextView {
```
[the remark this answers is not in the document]
nielsenko: Not sure I follow. 
```dart
xtension type AdminContextView(RequestContext _relicContext) { // missing implements
```
will introduce a compile error in:
```dart
FutureOr<Response> handleAdminDashboard(final AdminContextView context) {
  print(
      'Handling Admin Dashboard for ${context.user.name} (${context.adminRole.roleName})');
  return Response.ok(
      'Admin: ${context.user.name}, Role: ${context.adminRole.roleName}');
}
```
The meaning of implements for extension types are defined [here](https://dart.dev/language/extension-types#implements).

In particular:

> An extension type can only implement:
> ...
> - Another extension type that is valid on the same representation type. This allows you to reuse operations across multiple extension types (similar to multiple inheritance).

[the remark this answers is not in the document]
nielsenko: If I defined:
```dart
extension type AdminContextView._(RequestContext _relicContext)
    implements UserContextView {
  AdminContextView(UserContextView uv) : _relicContext = uv._relicContext;

  // Admin also has User
  AdminRole get adminRole => _adminRoleProperty.get(_relicContext);
  void attachAdminRole(final AdminRole role) =>
      _adminRoleProperty.set(_relicContext, role);
}
```
would be better?

In general the example don't do full encapsulation. You could argue the ctor should take `User` and do the _attach_.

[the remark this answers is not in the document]
tp: > Not sure I follow.

I did not mean removing the `implements` keyword, but rather the implementation (of assigning the user ID inside the admin context view).

I can remove the implementation, get it to compile and then error at runtime.
(We discussed this already, just for completeness sake here.)

`AdminContextView(UserContextView uv)` in the constructor seems like a better pattern to me, pushing the implementation to be more correct. Sadly that only solves 1 layer of inheritance. If you wanted to build a combined view that had `implements A, B` that would not work as well (unless you take 2 constructor params and hope that they point fo the same context…).

nielsenko:
A few bits survived, in particular `ContextProperty<T>`, but mostly this level of strict typing when composing middleware was deemed to cumbersome, and not attempted in earnest.

We may revisit the intention in this design later.

PR #117, opened 2025-07-14
fix: Update imports README.md

tp:
When running the example with version 0.4.1 I had to change the imports slightly from the suggested ones in the README.


nielsenko (approved):
LGTM 👍

tp:
@nielsenko didn't want to have you pinged during your time OOO, but feel free to merge this if it still applies :)

PR #161, opened 2025-09-09
fix: Set content length on body, not headers

nielsenko:
## Description
Content-Length is handled by `Body`. Don't set directly on `Headers`.

## Related Issues
Fixes: #138 

## Pre-Launch Checklist

## Breaking Changes
- [x] No breaking changes.

## Summary by CodeRabbit

- New Features
  - None.

- Bug Fixes
  - Corrected HTTP headers for static file responses: Content-Length is no longer sent in base responses where size may vary and is omitted for multipart/byteranges replies, improving standards compliance and client compatibility.

- Refactor
  - Internal adjustments to how response headers are built for range and multipart responses.

- Notes
  - No public API changes.

### DIFF
--- a/lib/src/io/static/static_handler.dart
+++ b/lib/src/io/static/static_handler.dart
@@ -213,7 +213,6 @@ Headers _buildBaseHeaders(
     final _FileInfo fileInfo, final CacheControlHeader? cacheControl) {
   return Headers.build((final mh) => mh
     ..acceptRanges = AcceptRangesHeader.bytes()
-    ..contentLength = fileInfo.stat.size
     ..etag = ETagHeader(value: fileInfo.etag)
     ..lastModified = fileInfo.stat.modified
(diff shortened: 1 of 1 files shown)


SandPod (approved):
Some background, this is automatically handled by the body right?

nielsenko:
> Some background, this is automatically handled by the body right?

Yes , but it is the adaptor that does it. We might want to move the reponsibility up the stack a bit.

Also - while it is a property closely related to the body of a response, it is a header, and perhaps it is more confusing that we let the `Body` "handle" it.

nielsenko:
> @nielsenko I think the thought was that it would be easy to missmatch the header with the body of these are configured in different places. I guess it is possible to override the value through the header if desired right? So you get a good resonable default but you can take control if you would like to.

Currently you can set it both ways, but if set on Body it will win.

PR #185, opened 2025-10-02
feat(trie): Add support for setting up mappings of values on lookup `trie.use(path, map)`

nielsenko:
## Description

Adds `PathTrie.use()` method to apply transformation functions to values on lookup. When a mapping is registered at a path, it transforms the value at that path and all descendant paths during lookup. Multiple mappings along a path compose from leaf to root.

Key changes:
- Added `map` field to `_TrieNode` to store transformation functions
- Implemented `use()` method to register mappings at specific paths
- Updated `lookup()` to compose and apply mappings during traversal
- Added conflict detection in `attach()` for overlapping maps
- Added test

The higher purpose for this change is to allow for setup of middleware at various points in the routing tree. PRs to update `Router<T>` as well as exposing in `Serverpod` will follow.

## Related Issues

- Fixes: #

## Pre-Launch Checklist

- [x] I have updated/added relevant documentation.

## Breaking Changes

- [x] No breaking changes.

## Summary by CodeRabbit

* New Features
  * Added per-path value transformation during route lookup, composing mappers along the matched path to modify returned values without changing routes.
* Refactor
  * Tightened generic bounds to non-nullable Object across routing types, requiring non-nullable type parameters in consumer code.
* Tests
  * Added comprehensive tests covering mapping on root, wildcards, parameters, and composition order.

### DIFF
--- a/test/router/path_trie_use_test.dart
+++ b/test/router/path_trie_use_test.dart
@@ -91,7 +91,7 @@ void main() {
   test(
       'Given multiple paths with values, '
       'when use is applied to a specific path, '
-      'then only that path and its descendants are transformed', () {
+      'then only descendants of that path are transformed', () {
     final trie = PathTrie<int>();
     final root = NormalizedPath('/');
     final pathA = NormalizedPath('/a');


nielsenko:
@coderabbit review

nielsenko:
The test coverage could be improved, but mostly for some error situations, such as setting up mappings on the same node twice, or attaching nodes to each other with conflicting mappings.

test/router/path_trie_test.dart
```diff
@@ -502,5 +502,119 @@ void main() {
         expect(result.parameters, equals({#id: '456'}));
       });
     });
+
+    group('Use mapping', () {
```
SandPod, line 506: Question: Should this group be represented by a file in a folder named `path_trie` instead? `path_trie_map_test.dart`, or perhaps even better `path_trie_use_test.dart` since that is actually the public interface.

Seems like this file is growing a bit and might need splitting out?

test/router/path_trie_test.dart
```diff
@@ -502,5 +502,119 @@ void main() {
...
+      test(
+          'Given a value at root, '
+          'when use is applied to root, '
+          'then the value is transformed', () {
+        final trie = PathTrie<int>();
+        final root = NormalizedPath('/');
+        trie.add(root, 1);
+        trie.use(root, (final i) => i * 2);
+        expect(trie.lookup(root)?.value, 2, reason: 'doubled');
```
SandPod, line 515: Suggestion: Not entirely sure how the reason is built but I believe this would give a better error message right?

```suggestion
        expect(trie.lookup(root)?.value, 2, reason: 'Value should have been doubled');
```

SandPod: Apply for the other tests with only `doubled` as well.

test/router/path_trie_test.dart
```diff
@@ -502,5 +502,119 @@ void main() {
...
+        expect(trie.lookup(pathA)?.value, 20, reason: 'doubled');
+        expect(trie.lookup(pathB)?.value, 200, reason: 'doubled');
+      });
+
+      test(
+          'Given multiple paths with values, '
+          'when use is applied to a specific path, '
+          'then only that path and its descendants are transformed', () {
```
SandPod, line 572: Suggestion(NIT): Perhaps separate tests with a shared setup? That would allow us to remove the `and` and be specific when we fail.

test/router/path_trie_test.dart
```diff
@@ -502,5 +502,119 @@ void main() {
...
+          trieA.lookup(NormalizedPath('/prefix/b'))?.value,
+          200,
+          reason: 'doubled',
+        );
+      });
+
+      test(
+          'Given multiple use mappings on ancestor and descendant paths, '
+          'when looking up a value, '
+          'then transformations are applied from leaf to root', () {
```
SandPod, line 609: Question: Wouldn't it be more logical if these where applied in the reverse order?

Say we attach a user object to the context in our root and then we want to do some additional information of that user object for a specific route. This order of operations wouldn't allow this. Now the root needs to be aware of the possible operations that the leaf performs.

lib/src/router/path_trie.dart
```diff
@@ -61,7 +66,7 @@ final class _Tail<T> extends _DynamicSegment<T> {}
 ///
 /// Supports literal segments, parameterized segments (e.g., `:id`), wildcard segments (`*`),
 /// and tail segments (`**`). Allows associating a value of type [T] with each complete path.
-final class PathTrie<T> {
+final class PathTrie<T extends Object> {
```
SandPod, line 69: Question(For my clarification): Is this needed now because we have assigned `null` a special meaning in our lookups?

test/router/path_trie_test.dart
nielsenko: We are not calling the handler during lookup, we are composing it to the be called after lookup. Middleware should wrap the sub-trie. 

So setup auth middleware on /authenticated and the info middleware at /authenticated/user. Then the auth will be called first doing some setup, delegate to its inner middleware info that will do some more setup, then delegate to some handler (perhaps), which returns a handled context to info, which may do something more to it, returning it to auth, finally passing it to relic.

That got a bit convoluted. Does it make sense?

test/router/path_trie_test.dart
nielsenko: You are probably correct. I should look it up though and try to memorize for the future.

nielsenko:
Changed semantics slightly in [24462c2](https://github.com/serverpod/relic/pull/185/commits/24462c2b0c7b7e9651a6dc0563168ac32e2ed19c). You are now allowed to call `use` on a hitherto non-existing path. Ie to call `use` before `add`.

Also extended test coverage a bit.

test/router/path_trie_test.dart
nielsenko: What about the slightly shorter `Should double`. Looks like this on failure:
```
00:00 +37 -1: PathTrie<T> Use mapping Given two tries where one is attached to the other, when use is applied to the root of the child, then only attached trie values are transformed [E]
  Expected: <200>
    Actual: <100>
  Should double
  
  package:matcher                        expect
  test/router/path_trie_test.dart 641:9  main.<fn>.<fn>.<fn>
  
```

nielsenko:
Changed semantics in [7e6e34a](https://github.com/serverpod/relic/pull/185/commits/7e6e34a64333e4547780d3de8f193ccb2de4d75c) to allow `use` to be called multiple times for the same path. This is also impacts `attach`.

test/router/path_trie_test.dart
nielsenko: Split out into `path_trie_use_test .dart`

test/router/path_trie_test.dart
nielsenko: Cheated a bit and got rid of the and.

I would prefer not to split this test-case in 3. It is just half a page of code.

PR #186, opened 2025-10-03
feat(router): Add support for setting up mappings of values on lookup `router.use(path, map)`

nielsenko:
## Description

Adds `Router.use()` method to support middleware-style transformations of routes. The method applies a mapping function to all routes under a given path prefix. Transformations compose hierarchically and are applied during lookup leaf-to-root, with execution order from root-to-leaf in the case of functions (such as `Handlers`).

Includes tests covering basic transformation, composition, multiple HTTP methods, attached routers, and hierarchical middleware. 

## Related Issues

Fixes: #71 

## Pre-Launch Checklist

## Breaking Changes

- [x] No breaking changes.

## Summary by CodeRabbit

* **New Features**
  * Adds route-level middleware allowing path-scoped transformations, composition of multiple middlewares, and hierarchical application across nested routers.
* **Tests**
  * Adds comprehensive unit tests validating middleware composition, ordering, propagation to descendants, and interaction with attached routers.
* **Examples**
  * Adds a runnable middleware example demonstrating auth middleware, request handling, and client-server interaction.

### DIFF
--- a/test/router/router_use_test.dart
+++ b/test/router/router_use_test.dart
@@ -1,9 +1,5 @@
-import 'package:http/http.dart' as http;
 import 'package:relic/relic.dart';
 import 'package:test/test.dart';
-
-import '../headers/headers_test_utils.dart';
-import '../util/test_util.dart';
 
 void main() {
   test(
@@ -155,54 +151,4 @@ void main() {
       '<a><b><c>request</c></b></a>',
     );
   });
-
-  test(
-      'Given authentication middleware, '
-      'when a handler is invoked, '
-      'then it can retrieve the user', () async {
-    final router = Router<Handler>();
-    router
-      ..use('/', AuthMiddleware().call)
-      ..get(
-          '/api/user/info',
-          (final ctx) =>
-              ctx.respond(Response.ok(body: Body.fromString('${ctx.user}'))));
-
-    final relic = await testServe(const Pipeline()
-        .addMiddleware(routeWith(router))
-        .addHandler(respondWith((final _) => Response.notFound())));
-
-    final response =
-        await http.get(relic.url.replace(path: '/api/user/info'), headers: {
-      'Authorization': 'Bearer 42', // just an example
-    });
-
-    expect(response.statusCode, equals(200));
-    expect(response.body, '42');
-  });
-}
-
-typedef User = int;
-final _auth = ContextProperty<User>('auth');
-
-extension on RequestContext {
-  User get user => _auth[this];
-}
-
-class AuthMiddleware {
-  bool _validate(final String token) => true; // just an example
-  User _extractUser(final String token) => int.parse(token);
-
-  Handler call(final Handler next) {
-    return (final ctx) {
-      final bearer =
-          ctx.request.headers.authorization as BearerAuthorizationHeader?;
-      if (bearer == null || !_validate(bearer.token)) {
-        return ctx.respond(Response.unauthorized());
-      } else {
-        _auth[ctx] = _extractUser(bearer.token);
-        return next(ctx);
-      }
-    };
-  }
 }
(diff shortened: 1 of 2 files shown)


nielsenko:
@coderabbit review

lib/src/router/router.dart
```diff
@@ -57,6 +57,23 @@ final class Router<T extends Object> {
     );
   }
 
+  /// Adds a middleware function to the router.
+  ///
+  /// The [path] defines the route pattern. Segments starting with `:` (e.g.,
+  /// `:id`) are treated as parameters. The associated [map] function is applied to
+  /// the matched route before executing the route handler.
```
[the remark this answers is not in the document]
nielsenko: It does not modify the stored value. The transformation is applied during lookup, but is never stored.

lib/src/router/router.dart
```diff
@@ -57,6 +57,23 @@ final class Router<T extends Object> {
     );
   }
 
+  /// Adds a middleware function to the router.
+  ///
+  /// The [path] defines the route pattern. Segments starting with `:` (e.g.,
+  /// `:id`) are treated as parameters. The [map] function transforms matched route
+  /// values during lookup, without modifying the stored routes.
+  void use(final String path, final T Function(T) map) {
+    // TODO: While correct, it would be enough to map only the entry
```
SandPod, line 66: Suggestion: Please remove this TOOD and add an issue for this.

This is an optimization I don't think needs to be recorded in the implementation. We should track it with issues.

test/router/router_use_test.dart
```diff
@@ -0,0 +1,156 @@
+import 'package:relic/src/router/lookup_result.dart';
+import 'package:relic/src/router/method.dart';
+import 'package:relic/src/router/router.dart';
+import 'package:test/test.dart';
+
+void main() {
```
SandPod, line 6: Question: Should we perhaps add an authenticated test for this as well, just to showcase how we envision this being used on a high level?

That is, adding an object to the context from the middleware that is attached to the router?

SandPod (commented):
Changes looks good, two questions.

lib/src/router/router.dart
```diff
@@ -57,6 +57,23 @@ final class Router<T extends Object> {
...
+    // for the method that is actually looked up. However this is not
+    // easily achieved with the current implementation.
```
nielsenko, line 68:
```suggestion
    // TODO: While correct, it would be enough to map only the entry
    // for the method that is actually looked up. However this is not
    // easily achieved with the current implementation.
```
@coderabbit Create an issue to track this potential optimization

test/router/router_use_test.dart
SandPod: Yes, I think an example might be better. Just to showcase the capability

nielsenko: Doh! Added it as a test 😄 .. Could you take a look at 
[2947c5f](https://github.com/serverpod/relic/pull/186/commits/2947c5fd5b6935368ae019c383b0f78110a8ce6d) before I rework it as an example - to see if this aligned with your expectations?

SandPod: Test looks good, but perhaps we could reflect the user in the header instead of having a magic value? That would make it even more explicit.

nielsenko: User id now fetch from Bearer token.

SandPod (approved):
LGTM! But should the test be reworked as an example?

nielsenko:
> LGTM! But should the test be reworked as an example?

Reworked as example in [8f367e7](https://github.com/serverpod/relic/pull/186/commits/8f367e70de8cadd5c32b05fad995ff724aafcc4f)

nielsenko:
@coderabbit review

PR #241, opened 2025-10-24
chore: Update expired certificates

nielsenko:
Update expired certificates in ssl_certs.dart


nielsenko:
This just rolls the certificates to something that is still valid. Will follow up with scripting to create cert chains.

PR #280, opened 2025-11-24
docs: Update code block references in documentation

Zfinix:
## Description
docs: Update code block references in documentation to point to the correct example files

## Breaking Changes
- [x] No breaking changes.


Zfinix:
> LGTM, but we should probably test everything locally before merging.

Sure

PR #284, opened 2025-11-25
docs: Improved README file.

vlidholt:
Updates the main README file for Pub.


Zfinix (approved):
it looks great but i think it might benefit from having the examples section here also and a note on how to run the examples

PR #298, opened 2025-12-03
test: Less delay when running locally than on CI

nielsenko:


nielsenko:
Until I get around to do a cross-isolate completer thingy..

PR #309, opened 2025-12-15
feat: Add optional consume flag to attach

nielsenko:
## Description

This PR makes it an explicit error to try to `attach` on a tail path (`../**`). Previously, this would fail in subtle ways because `attach` cannot really extend tail segments, even though it didn't fail hard. This in turn makes the one acceptable corner-case - attaching a router with a single route mounted at root - illegal.

To mitigate this, the PR adds a `consume` flag to `attach`. When `consume: true`, the sub-router is detached after attachment, allowing single-value tries to be safely mounted at tail paths.

This change enables `injectAt` to still work with tail paths (`/**`) when the `injectable` only adds a single route at `/`, which is useful for mounting simple handler objects with `injectAt`.

- Add `consume` flag to `attach` that resets the attached trie's root after attachment
- Add `isSingle` property to check if a trie/router has only a single root route
- Allow `attach(..., consume: true)` at tail paths (`/**`) when the attached trie is single-value
- Update `injectAt` to use `attach(..., consume: true)`, enabling injection at tail paths for simple `HandlerObject`s
- Prevent `attach` and `group` from being used with tail paths (unless `consume: true` with single-value trie)

## Related Issues
- Fixes: #

## Pre-Launch Checklist

- [x] I have updated/added relevant documentation.

## Breaking Changes

- `attach` now throws `ArgumentError` when called with a tail path (`/**`) unless `consume: true` and the attached trie is single-value
- `group` now throws `ArgumentError` when called with a tail path

## Summary by CodeRabbit

* **New Features**
  * Added `attachAndConsume` method to Router and RelicApp for consuming attached sub-routes in a single operation.
  * Added `isSingle` getter to Router and RelicApp to check if the router has a single root route.

* **Bug Fixes**
  * Added validation to prevent attaching sub-tries at tail paths, throwing an error for conflicting semantics.

* **Chores**
  * Updated minimum Dart SDK requirement from 3.7.0 to 3.8.0.
  * Updated dev dependencies and improved code formatting.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/lib/src/router/path_trie.dart
+++ b/lib/src/router/path_trie.dart
@@ -325,17 +325,19 @@ final class PathTrie<T extends Object> {
   /// For example, if this trie has a path `/a/b` and you attach another trie
   /// at `/a/b/c`, the attached trie's root will be accessible via `/a/b/c`.
   ///
+  /// If [consume] is `true`, the attached trie's root is reset after attachment,
+  /// preventing further modifications to the attached trie from affecting this
+  /// trie. This also allows attaching a single-value trie at a tail path (`/**`).
+  ///
   /// Throws an [ArgumentError] if:
   /// - The node at [normalizedPath] has a value, and the root node of [trie] has as well.
   /// - Both nodes has an associated dynamic segment.
   /// - There are overlapping children between the nodes.
-  void attach(final NormalizedPath normalizedPath, final PathTrie<T> trie) =>
-      _attach(normalizedPath, trie, consume: false);
-
-  void _attach(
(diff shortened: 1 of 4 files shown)


nielsenko:
@Sandpod To make pease with CI I updated SDK to 3.8 which causes yet some more formatting changes. You can disregard the two last commits. Or if you prefer I can split them out in a separate PR, and rebase this on top for a cleaner history.

SandPod (approved):
LGTM! 👍

PR #312, opened 2025-12-15
chore: Bump version to 0.14.0 and update CHANGELOG.md

nielsenko:
Prepare for release of 0.14.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,11 @@
 ## 0.14.0
 - feat!: Add optional consume flag to attach ([#309](https://github.com/serverpod/relic/pull/309))
   - BREAKING: `attach` now throws `ArgumentError` when called with a tail path (`/**`), unless `consume: true` and the attached trie/router has a single value at the root (`isSingle`) 
-  - BREAKING: `group` now throws `ArgumentError` when called with a tail path.
+  - BREAKING: `group` now throws `ArgumentError` when called with a tail path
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 0.14.0
+- feat!: Add optional consume flag to attach ([#309](https://github.com/serverpod/relic/pull/309))
+  - BREAKING: `attach` now throws `ArgumentError` when called with a tail path (`/**`), unless `consume: true` and the attached trie/router has a single value at the root (`isSingle`) 
+  - BREAKING: `group` now throws `ArgumentError` when called with a tail path.
+  - Add `consume` flag to `attach` that resets the attached trie's root after attachment
+  - Add `isSingle` property to check if a trie/router has only a single root value/route
+  - Allow `attach(..., consume: true)` at tail paths (`/**`) when the attached trie/router `isSingle`
+  - Update `injectAt` to use `attach(..., consume: true)`, still enabling injection at tail paths for simple `HandlerObject`s (those adding a single handler at the root ('/'))
+
+
```
marcelomendoncasoares: A few general comments:

- Some entries end with period, some don't - could be nice to standardize, regardless of which.
- The `unless` condition of the first entry uses features that are introduced below. Would it make sense to only reference them only (e.g. "see consume for how to disable") and explain the "unless" on the `consume` entry?
- Two newlines between the last entry and the previous version (different from previous spacing).

PR #343, opened 2026-03-10
feat: Make NormalizedPath interning cache configurable

### LINKED ISSUE #118: Improve interning of NormalizedPath
### Problem to Solve

The `NormalisedPath` class use interning to improve performance (In particular for `hashCode` and `equals`).

The current implementation uses a simple LRU cache, which works really well until it doesn't. You can construct input so that the hit-ratio approaches 0, in which case we are no longer the fastest trie on the block.

While such inputs are highly unlikely to happen in real use (ie. outside evil benchmarks), we should experiment with other cache designs. I expect we can improve not only worst case, but also common case performance.

Even if we stick with simple LRU, it should be possible to tune the size of the cache, to accommodate your workload.

### Proposal

- Try other cache designs.
- Perhaps allow user to inject what cache to use.
- At least allow LRU cache size to be tuned.

### Use Case

Caches helps performance, but can suffer from thrashing/performance cliffs. Allowing users to tune caching to their workload can help mitigate this.

### Alternatives

Wait and see, if this ever becomes a problem.

### Additional context

_No response_

### How experienced are you with this library?

Expert - Experienced and comfortable with using this library in complex projects

### Are you interested in working on a PR for this?

- [ ] I want to work on this

### LINKED ISSUE #342: Performance cliff on high-cardinality dynamic paths due to NormalizedPath interning
`Router.lookup(...)` always constructs a `NormalizedPath(path)`, and `NormalizedPath` internally uses a fixed-size LRU intern cache:

```dart
```

This is useful when full request paths repeat often, but it causes a clear performance cliff under high-cardinality dynamic path workloads.

Minimal reproduction shape:

- route pattern: `/users/:id`
- method: `GET`
- `50000` lookups
- consume match result and params length
- vary the number of unique full paths (`cardinality`)

Observed local results:

| cardinality | relic |
|---|---:|
| 1 | 7572us |
| 16 | 7218us |
| 256 | 7185us |
| 1024 | 7241us |
| 4096 | 7894us |
| 10000 | 72734us |
| 20000 | 257738us |
| 50000 | 252520us |

The pattern is:

- lookup stays relatively stable while the number of unique full paths remains below the intern-cache capacity
- once it reaches and exceeds that capacity, lookup time rises sharply

This is not a correctness bug, but it is a fairly obvious performance cliff.

For routes like `/users/:id`, if full paths have low reuse, interning can turn from an optimization into a liability.

Minimal reproduction code:

```dart
import 'package:relic/relic.dart' as relic;

const lookups = 50000;
const rounds = 7;
const cardinalities = [1, 16, 256, 1024, 4096, 10000, 20000, 50000];

int median(List<int> values) {
  final sorted = [...values]..sort();
  return sorted[sorted.length ~/ 2];
}

List<String> buildPaths(int cardinality) {
  return List<String>.generate(
    lookups,
    (i) => '/users/user_${i % cardinality}',
  );
}

int runRelic(List<String> paths) {
  final router = relic.Router<int>()..get('/users/:id', 1);
  var sink = 0;
  final sw = Stopwatch()..start();
  for (final path in paths) {
    final match = router.lookup(relic.Method.get, path).asMatch;
    sink ^= match.value;
    sink ^= match.parameters.length;
  }
  sw.stop();
  if (sink == -1) throw StateError('unreachable');
  return sw.elapsedMicroseconds;
}

void main() {
  print('lookups=$lookups');
  print('rounds=$rounds');
  print('pattern=/users/:id');
  print('lower is better (us)');

  for (final cardinality in cardinalities) {
    final paths = buildPaths(cardinality);
    final samples = <int>[];
    for (var round = 0; round < rounds; round++) {
      samples.add(runRelic(paths));
    }
    print('cardinality=$cardinality relic=${median(samples)}');
  }
}
```

Possible directions:

1. make `NormalizedPath` interning configurable or disable-able
2. reduce cache churn cost for low-reuse workloads
3. document this performance boundary explicitly

nielsenko:
Introduce abstract Cache<K, V> interface with LruCache and NoCache implementations. NormalizedPath.interned can now be swapped to tune caching strategy for different workloads.

Fixes #342, fixes #118

## Release Notes

* **New Features**
  * Introduced configurable caching strategies for route normalization: choose between LRU caching for performance or opt out of caching for high-cardinality workloads.

* **Tests**
  * Added comprehensive tests validating caching behavior and path normalization across different strategies.


nielsenko:
@coderabbit review

nielsenko:
Please disregard coverage drop. Dart doesn't actually call an empty ctor. It only exist to mark it `const`.

image

nielsenko:
While this makes caching configurable, we should investigate if we can make no-interning even cheaper. Dart side caches cannot be shared across isolates, which makes them less useful in multi-isolate deployments.

@coderabbit Please create a follow-up issue.

marcelomendoncasoares (approved):
LGTM!

PR #345, opened 2026-03-11
chore: Format snippets

nielsenko:
- Use snip tool to format dart code in markdown and dart doc comments. 
- Some corrections where needed as not all snippets where valid dart code. 
- Also adds check on ci.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,7 +44,10 @@ jobs:
         run: dart pub get
 
       - name: Verify Formatting
-        run: dart format --output=none --set-exit-if-changed .
+        run: >-
+          dart format --output=none --set-exit-if-changed . &&
+          dart pub global activate --source git https://github.com/nielsenko/snip.git &&
+          snip format . --no-apply
 
       - name: Analyze (downgraded)
         run: >-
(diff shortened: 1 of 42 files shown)


.github/workflows/ci.yaml
```diff
@@ -44,7 +44,10 @@ jobs:
         run: dart pub get
 
       - name: Verify Formatting
-        run: dart format --output=none --set-exit-if-changed .
+        run: >-
+          dart format --output=none --set-exit-if-changed . &&
+          dart pub global activate --source git https://github.com/nielsenko/snip.git &&
+          snip format . --no-apply
```
marcelomendoncasoares, line 50: Question: Is it possible to preserver trailing commas on the formatting? I would say that some of the snippets lost a bit in readability due to not preserving them.

marcelomendoncasoares (commented):
Great UX with the `snip` tool!! Next step for the future is to have it in a pre-commit hook so we ensure to never have conflicts on CI. I have only one question, though, that I think would be nice to preserve on the snippets as they were.

.github/workflows/ci.yaml
nielsenko: Yes.. 2m

Issue #113, opened 2025-07-02
Overhaul typed header parsers


nielsenko:
While the idea of typed headers is great, we are not particular compliant to the various RFCs when it comes to parsing headers.

We should use a composable parser library such [petitparser](https://pub.dev/packages/petitparser) and implement the various header parsers by composing the parsers in strict accordance with the RFCs. 

Go over all the (many) typed header parses and ensure they are compliant with their respective RFCs.

nielsenko:
Did a pass over all 45 typed headers under `packages/relic_core/lib/src/headers/typed/headers/` against their RFCs / defacto specs. Findings below as a concrete checklist to drive the overhaul.

Out of scope (tracked separately, not repeated below):
- Quoted-string-aware splitting on `;`/`=`/`,` is covered by #102. This touches `forwarded_header.dart`, the `;`-split parts of `content_disposition_header.dart`, etc.
- `SetCookieHeader` `Domain`/`Path` typed as `Uri` and serialized with `Uri.toString()` is filed as #355.

Line references are from `main` at the time of audit and may drift. Use them as starting points, not anchors.

## Critical: encoder produces invalid wire output

- [ ] `content_range_header.dart` (RFC 9110 14.4): encoder emits literal `"bytes null-499/1234"` when `start==null && end!=null` via Dart string interpolation. Unsatisfied range with `size==null` emits `bytes */*`; spec requires `*/complete-length`.
- [ ] `access_control_allow_origin_header.dart` (Fetch 3.2.2): stores origin as `Uri`, so `Uri.parse("https://example.com/")` round-trips with the trailing slash. Same shape bug as #355. Also `null` sentinel not modeled (lands as a degenerate `Uri` with scheme="" path="null"; happens to encode "null" by accident).
- [ ] `host_header.dart` (RFC 9110 7.2): `HostHeader.fromUri` uses `uri.port`, which returns the default port (80/443) when no port was specified. Result: `Host: example.com:80`. Fix: `uri.hasPort ? uri.port : null`.
- [ ] `authorization_header.dart` Digest encoder (RFC 7616 3.4): always quotes every parameter; `algorithm`, `qop`, `nc`, `stale` MUST be unquoted tokens. Strict servers (Apache mod_auth_digest) reject.
[...]

PR #361, opened 2026-06-23
refactor: Cookie and Set-Cookie header

nielsenko:
## Description
Hardens cookie header handling and reshapes the Set-Cookie types so the single-cookie value and the header collection are distinct, matching the request side.

- Rename the single-cookie class `SetCookieHeader` -> `SetCookie` (name/value + attributes)
- `SetCookieHeader` is now a `List<SetCookie>` collection, one element per `Set-Cookie` response line 
  (never comma-folded, RFC 6265 4.1).
- `SetCookieHeader` gains `.empty()`, `add`, `addAll`, value equality, and a multi-value codec
- `SetCookie.encode()` is now public.
- `CookieHeader.parse` is lenient: skip malformed cookies and keep the well-formed ones.
  One bad third-party cookie no longer hides a valid session/auth cookie. Throw only when no usable cookie remains. 
- `SetCookieHeader.parse` stays strict (server-produced, no skip rationale).
- Add `CookieHeader.getCookies(name)` returning every match in header order.
- `getCookie` returns the first of possible duplicates (RFC 6265 5.4).
- `SetCookie` strips a leading dot from the `Domain` (RFC 6265 5.2.3).

## Related Issues
- Related: #113 

## Pre-Launch Checklist
- [x] I have read and followed the Dart Style Guide and formatted with dart format.
- [x] I have updated/added relevant documentation (doc comments with `///`).

## Breaking Changes
- [ ] No breaking changes.

### DIFF
--- a/packages/relic/test/headers/typed/cookie_header_test.dart
+++ b/packages/relic/test/headers/typed/cookie_header_test.dart
@@ -135,8 +135,8 @@ void main() {
       },
     );
 
-    test('when a valid Cookie header is passed with an empty name '
-        'then it should parse the cookies correctly', () async {
+    test('when a Cookie header carries a nameless `=value` segment '
+        'then it is dropped and the named cookies are kept', () async {
       final headers = await getServerRequestHeaders(
         server: server,
         touchHeaders: (final h) => h.cookie,
@@ -145,11 +145,11 @@ void main() {
 
       expect(
         headers.cookie?.cookies.map((final c) => c.name).toList(),
-        equals(['', 'userId']),
+        equals(['userId']),
       );
       expect(
         headers.cookie?.cookies.map((final c) => c.value).toList(),
-        equals(['abc123', '42']),
+        equals(['42']),
       );
     });
 
--- a/packages/relic/test/headers/typed/cookie_header_test.dart
+++ b/packages/relic/test/headers/typed/cookie_header_test.dart
@@ -36,20 +36,85 @@ void main() {
     );
 
     test(
-      'when a Cookie header with invalid format is passed then the server responds '
-      'with a bad request including a message that states the cookie format is invalid',
+      'when a cookie with an invalid format (no "=") is present '
+      'then it is skipped and the remaining valid cookies are still parsed',
+      () async {
+        // A single malformed cookie must not make the other cookies in the
+        // same header unreadable.
+        final headers = await getServerRequestHeaders(
+          server: server,
+          touchHeaders: (final h) => h.cookie,
+          headers: {'cookie': 'sessionId=abc123; invalidCookie'},
+        );
+
+        expect(
+          headers.cookie?.cookies.map((final c) => c.name).toList(),
+          equals(['sessionId']),
+        );
+        expect(
+          headers.cookie?.cookies.map((final c) => c.value).toList(),
+          equals(['abc123']),
+        );
+      },
+    );
+
+    test(
+      'when a cookie with an invalid name is present '
+      'then it is skipped and the remaining valid cookies are still parsed',
+      () async {
+        final headers = await getServerRequestHeaders(
+          server: server,
+          touchHeaders: (final h) => h.cookie,
+          headers: {'cookie': 'invalid name=abc123; userId=42'},
+        );
+
+        expect(
+          headers.cookie?.cookies.map((final c) => c.name).toList(),
+          equals(['userId']),
+        );
+        expect(
+          headers.cookie?.cookies.map((final c) => c.value).toList(),
+          equals(['42']),
+        );
+      },
+    );
+
+    test(
+      'when a cookie with an invalid value is present '
+      'then it is skipped and the remaining valid cookies are still parsed',
+      () async {
+        final headers = await getServerRequestHeaders(
+          server: server,
+          touchHeaders: (final h) => h.cookie,
+          headers: {'cookie': 'sessionId=abc123; userId=42\x7F'},
+        );
+
+        expect(
+          headers.cookie?.cookies.map((final c) => c.name).toList(),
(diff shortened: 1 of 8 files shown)


packages/relic/test/headers/typed/cookie_header_test.dart
```diff
@@ -35,69 +35,83 @@ void main() {
...
-              (final e) => e.message,
-              'message',
-              contains('Invalid cookie format'),
-            ),
-          ),
+        final headers = await getServerRequestHeaders(
+          server: server,
+          touchHeaders: (final h) => h.cookie,
+          headers: {'cookie': 'sessionId=abc123; userId=42\x7F'},
         );
```
marcelomendoncasoares: Suggestion: Also test invalid before valid to ensure skip is consistent regardless of order.

packages/relic_core/test/headers/typed/set_cookie_header_test.dart
```diff
@@ -124,16 +136,64 @@ void main() {
...
+      });
+    });
+
+    group('Given an existing collection,', () {
+      test('when add is called, '
+          'then a new collection with the cookie appended is returned.', () {
+        final header = const SetCookieHeader.empty().add(a).add(b);
+
+        expect(header.cookies, equals([a, b]));
+      });
```
marcelomendoncasoares, line 336: Question: Should we also test that `empty()` returns a `const` list that is not mutated from `add`/`addAll`?

packages/relic_core/lib/src/headers/typed/headers/set_cookie_header.dart
```diff
@@ -242,9 +251,67 @@ final class SetCookieHeader {
...
+      value.cookies.map((final c) => c.encode()).toList();
+
+  @override
+  bool operator ==(final Object other) =>
+      identical(this, other) ||
+      other is SetCookieHeader &&
+          const ListEquality<SetCookie>().equals(cookies, other.cookies);
+
+  @override
+  int get hashCode => const ListEquality<SetCookie>().hash(cookies);
```
marcelomendoncasoares, line 318: Question: Are we testing equality? Did not see on the tests added on this PR, but might exist on previously existing ones.

packages/relic_core/lib/src/headers/typed/headers/set_cookie_header.dart
```diff
@@ -258,7 +325,7 @@ final class SameSite {
 
   /// Cookie with this value will be sent in all requests.
   ///
-  /// [Cookie.secure] must also be set to true, otherwise the `none` value
+  /// [SetCookie.secure] must also be set to true, otherwise the `none` value
   /// will have no effect.
   static const none = SameSite._('None');
```
marcelomendoncasoares, line 342: Question: Should we have a test for `SetCookie` with `secure: false`?

packages/relic_core/lib/src/headers/typed/headers/cookie_header.dart
```diff
@@ -28,24 +28,58 @@ final class CookieHeader {
...
+    // path-scoped duplicates).
+    //
+    // The server cannot tell them apart from the header alone. Keep such
+    // same-name cookies rather than rejecting the whole header.
+    // splitTrimAndFilterUnique still collapses byte-identical segments, which
+    // is harmless since they are indistinguishable. [getCookie] returns the
+    // first match.
+    //
+    // Cookie names are case-sensitive per RFC 6265 4.2.2/5.4, so `Sid` and
+    // `sid` stay distinct.
```
marcelomendoncasoares, line 48: Question: Are we validating this on tests?

packages/relic/test/headers/typed/cookie_header_test.dart
marcelomendoncasoares, line 20: Discussion: These tests are a bit out of the pattern because the group is rather generic and the test cases are described on the `when` clauses. Commenting here, but valid for the changes on the tests under this group. Since we're touching the cooking implementation, could be a good time to reorganize all at once. The correct would be the cookie description on the `Given` (i.e. "Given an empty cookie header, when passed to the server, then it responds...").

packages/relic_core/test/headers/typed/set_cookie_header_test.dart
```diff
@@ -124,16 +136,64 @@ void main() {
...
+    group('Given a malformed Set-Cookie line,', () {
+      test('when decoded, '
+          'then it throws (strict decode)', () {
+        // Set-Cookie is server-produced, so there is no need to be lenient
+        expect(
+          () => SetCookieHeader.codec.decode(['a=1', 'no-equals', 'b=2']),
+          throwsFormatException,
+        );
+      });
+    });
```
marcelomendoncasoares, line 372: Suggestion: Collapse these single-test groups into a `test` only, without the group.

packages/relic_core/test/headers/typed/set_cookie_header_test.dart
```diff
@@ -124,16 +136,64 @@ void main() {
...
+        final header = const SetCookieHeader.empty().add(a).add(b);
+
+        expect(header.cookies, equals([a, b]));
+      });
+
+      test('when addAll is called, '
+          'then all cookies are appended in order.', () {
+        expect(SetCookieHeader([a]).addAll([b]).cookies, equals([a, b]));
+      });
+    });
```
marcelomendoncasoares, line 342: Suggestion: Move the `existing collection` to the `setUp` of the "Given" group instead of declaring inside each test.

marcelomendoncasoares (changes requested):
A few suggestions for coverage and improving the pattern on the tests.

marcelomendoncasoares (approved):
LGTM! 🚀

PR #369, opened 2026-08-10
fix: Address security audit findings

nielsenko:
## Description
Fixes from the security review of relic: request parsing and header hardening, shutdown robustness for detached (hijacked/upgraded) connections. Non are deemed critical enough to warrant a GHSA (or CVE).

- `fix`: Apply tail node middleware when the tail matches empty.
- `fix`: Parse auth headers with the linear header scanner (single pass, junk between params rejected, no quadratic backtracking).
- `fix`: Stream multi-range responses lazily and cap the range count.
- `fix`: Draw the multipart boundary from a cryptographic generator.
- `fix`: Await the adapter dispatch so a response that fails to write yields a 500 instead of a pinned connection.
- `fix`: Validate header names and values on assignment (CR/LF/NUL and invalid names rejected at `Headers.build`).
- `docs`: Remove the unimplemented hidden file protection claim.
- `fix`: Normalize the path before prepending the routing host.
- `fix`: Split the request path into segments before decoding, so encoded slashes cannot alter routing.
- `feat`: Reject cross-origin WebSocket upgrades by default (`allowAnyOrigin` opts out).
- `fix`: Escape Content-Disposition parameter values.
- `fix`: Validate and escape AuthenticationHeader challenge parts.
- `fix`: Enforce the token grammar on MIME type parts, closing header injection via the content type at write time.
`handlerCompleted` guard).
- `fix`: Track and close detached connections on shutdown: counted in `connectionsInfo`, sent 1001 and drained on graceful close, destroyed without awaiting peers on forced close.

## Related Issues
- Related: internal security review (no public issue)

## Pre-Launch Checklist
- [ ] This update focuses on a single feature or bug fix. (For multiple fixes, please submit separate PRs.)

## Breaking Changes
- Cross-origin WebSocket upgrades are rejected by default; pass `WebSocketUpgrade(..., allowAnyOrigin: true)` to restore the old behavior.
- Invalid header names and values (including CR/LF/NUL) now throw on assignment instead of being emitted.
- Malformed auth headers and non-token MIME type parts that were previously tolerated are now rejected.

## Additional Notes
The same fixes are backported to the 1.x line in a #371 .


packages/relic_core/lib/src/headers/typed/headers/authentication_header.dart
```diff
@@ -44,33 +45,44 @@ final class AuthenticationHeader {
...
+
   /// Converts the [AuthenticationHeader] instance into a string representation
   /// suitable for HTTP headers.
   String _encode() {
+    Token.validate(scheme);
     final paramsString = parameters
         .map((final param) {
-          if (param.key.isEmpty) return param.value;
-          return '${param.key}="${param.value}"';
+          if (param.key.isEmpty) return Token.validate(param.value);
```
FXschwartz: This breaks round trip for valid `token68` values. `token68` allows `+`, `/` and trailing `=` (base64 padding), none of which are tchars:
```dart
final h = AuthenticationHeader.parse('Negotiate YWJjZA==');
AuthenticationHeader.codec.encode(h); // FormatException: Not a valid HTTP token
```
`parse` accepts it, `encode` throws. I think this branch needs a `token68` validator instead of `Token.validate`. The round-trip test only covers `abc123`, so the padding case slips through.

packages/relic_io/lib/src/adapter/io_relic_web_socket.dart
```diff
@@ -117,6 +117,33 @@ class IORelicWebSocket implements RelicWebSocket {
...
+  }
+
+  /// Terminates the connection without waiting for the peer.
+  ///
+  /// The adapter's shutdown hook, deliberately not part of [RelicWebSocket]:
+  /// applications close with [close]/[tryClose]; only the owner tearing the
+  /// server down needs a verb that makes progress without the peer's
+  /// cooperation. Callable regardless of [isClosed]; repeated calls have no
+  /// effect.
+  void destroy() {
```
FXschwartz: I feel like the doc and method name is a bit incorrect here. Since this doesn't really destroy anything and is just `closeGoingAway()` without the `await`, and dart:io still runs the close handshake. Could just be `unawaited(closeGoingAway())` with the broader catch instead of duplicating the close.

packages/relic_core/lib/src/router/normalized_path.dart
```diff
@@ -56,6 +56,36 @@ class NormalizedPath {
...
   }
 
+  /// Creates a [NormalizedPath] from segments that have already been split.
+  ///
+  /// Each element of [segments] is one segment and is never split further, so
+  /// a separator inside a segment stays part of it rather than introducing a
+  /// new one. Empty and `.` segments are dropped and `..` segments resolved,
+  /// exactly as for a path string, so the result carries the same guarantee
+  /// that no segment is `..`.
+  factory NormalizedPath.fromSegments(final Iterable<String> segments) {
```
FXschwartz: This loop is `_normalize` minus the split. Could `_normalize` could just feed `path.split('/')` through this? Also, `fromSegments` skips the interning cache, and via `fromUri` it's now on the hot path in the routing middleware, so every request looks like allocates a fresh instance while the class doc still claims instances are interned.

packages/relic_core/lib/src/headers/typed/headers/authorization_header.dart
```diff
@@ -49,6 +50,16 @@ abstract class AuthorizationHeader {
   }
 }
 
+/// Strips a case-insensitive auth-scheme [prefix] from [value] if present,
+/// returning the trimmed remainder, or [value] itself when absent.
+String _stripOptionalScheme(final String value, final String prefix) {
```
FXschwartz, line 55: Same prefix test as `_stripScheme`. Could `_stripScheme` sit on top of this? Strip, throw if nothing came off (identical(result, value))?

FXschwartz (changes requested):

packages/relic_core/lib/src/headers/typed/headers/authentication_header.dart
nielsenko: Good catch! token68 is `1*( ALPHA / DIGIT / "-" / "." / "_" / "~" / "+" / "/" ) *"="`. Using `Token.validate` is too simple.

packages/relic_io/lib/src/adapter/io_relic_web_socket.dart
nielsenko: .. yes you are right.. I really need to thread the socket through. 

Or perhaps just rename the method to something less pretentious 🤔

packages/relic_core/lib/src/router/normalized_path.dart
nielsenko: Yes, with a twist. Good catch on the interning.

packages/relic_io/lib/src/adapter/io_relic_web_socket.dart
nielsenko: Problem with calling actual `_socket.destroy()` is we would need to use `WebSocket.fromUpgradedSocket` to get a handle on `_socket` in the first place, and that require us to handle stuff like negotiating per-message-deflate etc. Or not support compression 😱 .

So, I'll just get rid of the badly named method.

nielsenko: Got rid of `destroy`. Call sites now just do `unawaited(closeGoingAway())` directly. `closeGoingAway` now swallows all exceptions, not just `io.WebSocketException`.

See [1e22521](https://github.com/serverpod/relic/pull/369/commits/1e22521e14420cec595c3de33a6a76206b0b51dd)

packages/relic_core/lib/src/router/normalized_path.dart
nielsenko: Fixed in [440fda4](https://github.com/serverpod/relic/pull/369/commits/440fda42af58a4bef4db406ba3017a4a0851d29f)

packages/relic_core/lib/src/headers/typed/headers/authorization_header.dart
nielsenko: Fixed in [e336526](https://github.com/serverpod/relic/pull/369/commits/e33652679d06161e610606f7bd96cf57fe91be3d)

packages/relic_core/lib/src/headers/typed/headers/authentication_header.dart
nielsenko: Fixed in [322081b](https://github.com/serverpod/relic/pull/369/commits/322081bce6a94c87a475fb845a37841401b42265)

packages/relic_io/lib/src/adapter/io_relic_web_socket.dart
nielsenko: It still skips interning. Will need to think a bit about this, but I don't think that should hold up the PR.

FXschwartz (approved):
LGTM!

nielsenko:
Folded fixups before merging

PR #371, opened 2026-08-10
fix: Address security audit findings (1.2)

nielsenko:
Backport of #369


packages/relic_core/lib/src/headers/typed/primitives/header_scanner.dart
```diff
@@ -0,0 +1,233 @@
+import 'token.dart';
```
FXschwartz, line 1: The 2.x tests for these should also be added right? main has `header_scanner_test.dart`, `token_test.dart` and `parameter_value_test.dart` under `test/headers/typed/primitives/`. The lib files look identical so the tests should be abled to be droped in as-is.

FXschwartz (changes requested):
Looking good! Just a handful of questions/comments:
1. No version bump and no CHANGELOG unless you were planning on doing that separately? This does have three breaking changes which is unfortunate but are we planning on just bumping to 1.3.0?
2. The same review comments I left on #369 apply here since the files are the same. Mainly the `token68` encode regression. Assuming those get fixed in #369 and re-backported rather than fixed twice.

nielsenko:
Backported fixups from #369. See 4c1c8c5d, 00fce924, and a5e339b7

nielsenko:
> 1. No version bump and no CHANGELOG unless you were planning on doing that separately? This does have three breaking changes which is unfortunate but are we planning on just bumping to 1.3.0?

Yes this ships as 1.3. Changelog will be separate commit. Regarding the breaking changes. This is correct, but doing breaking changes to patch-lines is preferable to leaving sec findings unaddressed. Also, the findings will not break compiles, only someone mistakenly depending on the wrong semantics. The weird `.validate` was done exactly to avoid making a const ctor non-const.

> 2. The same review comments I left on #369 apply here since the files are the same. Mainly the `token68` encode regression. Assuming those get fixed in #369 and re-backported rather than fixed twice.

Yes, backported.

packages/relic_core/lib/src/headers/typed/primitives/header_scanner.dart
nielsenko: I can do that yes, but they (mostly) come from main not #369

nielsenko: Fixed with 7568046, and  79cb220

FXschwartz (approved):
LGTM!

nielsenko:
Folded fixups before merging

PR #375, opened 2026-08-20
refactor: No normalized path interning

### LINKED ISSUE #118: Improve interning of NormalizedPath

### LINKED ISSUE #342: Performance cliff on high-cardinality dynamic paths due to NormalizedPath interning

### LINKED ISSUE #344: Investigate cheaper no-interning and multi-isolate cache sharing for NormalizedPath
## Background

PR #343 makes the `NormalizedPath` interning cache configurable (via the new `Cache<K, V>` interface, `LruCache`, and `NoCache` implementations). While this addresses the performance cliff identified in #342 and #118, a follow-up investigation is needed.

## Problem

- The current approach (even with `NoCache`) may not be as cheap as possible when interning is disabled. There may be room to make the no-interning path even lighter.
- Dart-side caches (including `LruCache`) **cannot be shared across isolates**. This significantly limits their effectiveness in multi-isolate deployments, where each isolate maintains its own separate cache with no shared benefit.

## Tasks

- [ ] Investigate whether the no-interning (`NoCache`) code path can be made even cheaper (e.g., avoiding unnecessary allocations or lookups entirely).
- [ ] Investigate options for cache sharing or coordination across isolates (e.g., using `SendPort`/`ReceivePort`, native code, or alternative architectural approaches).
- [ ] Document the multi-isolate limitation clearly in the public API if no sharing solution is feasible.

## References

- PR: https://github.com/serverpod/relic/pull/343
- Comment: https://github.com/serverpod/relic/pull/343#issuecomment-$(gh pr view 343 --repo serverpod/relic --json comments --jq '.comments[-1].id' 2>/dev/null || echo 'see-pr-343')
- Fixes #342, #118

## Requested by

@nielsenko

nielsenko:
## Description

Remove NormalizedPath interning

This PR removes NormalizedPath interning completely.

Interning was originally added to avoid re-normalizing paths, but it always had two problems:

- A rather bad perf cliff: some inputs thrash the cache (#342).
- Per-isolate memory use: Dart makes it impossible (or very difficult) to share caches across isolates (#344).

On top of that, the new `NormalizedPath.fromSegments` and `NormalizedPath.fromUri` constructors - now the ones actually used on the request path - made interning tricky, so the security PRs (#369 & #371) simply skipped it for those paths.

Most of the lost performance is recouped by:

- Pre-computing repeated work.
- Extracting parameters into a single mutable map with undo-in-place during backtracking. Previously a copy was made at each parameter level so backtracking could just throw it away, but undoing in place avoids the copy and the allocation, reducing GC pressure.
- ~~Inferring from the registered routes whether backtracking is needed at all, and taking a faster non-backtracking walk when it isn't.~~

Results
```
┌────────┬─────────┬─────────┬────────────┐
│ routes │ Router  │ Spanner │ Routingkit │
├────────┼─────────┼─────────┼────────────┤
│ 1M     │ 1229 ns │ 1393 ns │ 3468 ns    │
├────────┼─────────┼─────────┼────────────┤
│ 100k   │ 994 ns  │ 1015 ns │ 3194 ns    │
├────────┼─────────┼─────────┼────────────┤
│ 10k    │ 579 ns  │ 402 ns  │ 1889 ns    │
└────────┴─────────┴─────────┴────────────┘
(measured on a 2.4 GHz 8-core Intel Core i9)
```

Fixes: #118 (not by improving caching, but by dropping it)
Fixes: #342 (this time by disabling interning altogether, instead of making caching opt-out)
Closes: #344 (investigation done)

## Breaking changes

- `NormalizedPath.interned` is removed

### DIFF
--- a/packages/relic_core/lib/src/router/no_cache.dart
+++ b/packages/relic_core/lib/src/router/no_cache.dart
@@ -1,14 +1,7 @@
 import 'cache.dart';
 
-/// A no-op [Cache] implementation that never stores or retrieves values.
-///
-/// Useful for high-cardinality workloads where caching causes more overhead
-/// than it saves (e.g., many unique dynamic paths like `/users/:id`).
-///
-/// Example:
-/// ```dart
-/// NormalizedPath.interned = NoCache();
-/// ```
+/// A no-op [Cache] that never stores or retrieves values, for opting out of
+/// caching in high-cardinality workloads where it costs more than it saves.
 final class NoCache<K, V> implements Cache<K, V> {
   /// Creates a no-op cache.
   const NoCache();
--- a/packages/relic_core/lib/src/router/path_trie.dart
+++ b/packages/relic_core/lib/src/router/path_trie.dart
@@ -60,7 +60,10 @@ sealed class _DynamicSegment<T> {
 final class _Parameter<T> extends _DynamicSegment<T> {
   final String name;
 
-  _Parameter(this.name);
+  /// The parameter [name] as a [Symbol], precomputed for use during lookup.
+  final Symbol symbol;
+
+  _Parameter(this.name) : symbol = Symbol(name);
 }
 
 final class _Wildcard<T> extends _DynamicSegment<T> {}
@@ -392,7 +395,13 @@ final class PathTrie<T extends Object> {
     final bool backtrack = true,
   }) {
     return backtrack
-        ? _lookupRecursive(_root, normalizedPath, 0, _root.map, const {})
+        ? _lookupRecursive(
+            _root,
+            normalizedPath,
+            0,
+            _root.map,
+            <Symbol, String>{},
+          )
         : _lookupIterative(normalizedPath);
   }
 
@@ -432,17 +441,14 @@ final class PathTrie<T extends Object> {
 
     final segment = segments[index];
 
+    TrieMatch<T>? next(_TrieNode<T> node, final T Function(T)? map) =>
+        _lookupRecursive(node, normalizedPath, index + 1, map, parameters);
+
     // Try literal match first
     final child = node.children[segment];
     if (child != null) {
       final newMap = _composeMap(currentMap, child.map);
-      final result = _lookupRecursive(
-        child,
-        normalizedPath,
-        index + 1,
-        newMap,
-        parameters,
-      );
+      final result = next(child, newMap);
       if (result != null) return result;
       // Fall through to try dynamic segment
     }
@@ -452,9 +458,6 @@ final class PathTrie<T extends Object> {
     if (dynamicSegment != null) {
       final dynamicNode = dynamicSegment.node;
       final newMap = _composeMap(currentMap, dynamicNode.map);
-      final newParams = dynamicSegment is _Parameter<T>
-          ? {...parameters, Symbol(dynamicSegment.name): segment}
-          : parameters;
 
       if (dynamicSegment is _Tail<T>) {
         // Tail matches: check for value at this position
@@ -463,19 +466,18 @@ final class PathTrie<T extends Object> {
           value = newMap?.call(value) ?? value;
           return TrieMatch(
             value,
-            newParams,
+            parameters,
             normalizedPath.subPath(0, index),
             normalizedPath.subPath(index),
           );
         }
+      } else if (dynamicSegment is _Parameter<T>) {
+        parameters[dynamicSegment.symbol] = segment;
+        final result = next(dynamicNode, newMap);
+        if (result != null) return result;
+        parameters.remove(dynamicSegment.symbol); // backtrack
       } else {
-        return _lookupRecursive(
-          dynamicNode,
-          normalizedPath,
-          index + 1,
-          newMap,
-          newParams,
-        );
+        return next(dynamicNode, newMap);
       }
     }
 
@@ -524,7 +526,7 @@ final class PathTrie<T extends Object> {
         currentNode = dynamicSegment.node;
         updateMap();
         if (dynamicSegment case final _Parameter<T> parameter) {
-          parameters[Symbol(parameter.name)] = segment;
+          parameters[parameter.symbol] = segment;
         }
         if (dynamicSegment is _Tail<T>) break; // possible early match
       }
--- a/packages/relic_core/lib/src/router/path_trie.dart
+++ b/packages/relic_core/lib/src/router/path_trie.dart
@@ -472,10 +472,15 @@ final class PathTrie<T extends Object> {
           );
         }
       } else if (dynamicSegment is _Parameter<T>) {
+        final prev = parameters[dynamicSegment.symbol];
         parameters[dynamicSegment.symbol] = segment;
         final result = next(dynamicNode, newMap);
         if (result != null) return result;
-        parameters.remove(dynamicSegment.symbol); // backtrack
+        if (prev == null) {
+          parameters.remove(dynamicSegment.symbol);
+        } else {
+          parameters[dynamicSegment.symbol] = prev;
+        }
       } else {
         return next(dynamicNode, newMap);
       }
--- a/packages/relic_core/test/router/no_cache_test.dart
+++ b/packages/relic_core/test/router/no_cache_test.dart
@@ -28,68 +28,32 @@ void main() {
     });
   });
 
-  group('Given NormalizedPath with NoCache', () {
-    late Cache<String, NormalizedPath> originalCache;
+  group('Given a Router (no result cache)', () {
+    test('when a request path is looked up twice '
+        'then routing works and the results are equal but not memoized', () {
+      final router = Router<int>()..get('/a/b', 1);
 
-    setUp(() {
-      originalCache = NormalizedPath.interned;
-      NormalizedPath.interned = const NoCache();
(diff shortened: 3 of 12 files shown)


packages/relic_core/lib/src/router/path_trie.dart
```diff
@@ -463,19 +487,18 @@ final class PathTrie<T extends Object> {
...
+            parameters,
             normalizedPath.subPath(0, index),
             normalizedPath.subPath(index),
           );
         }
+      } else if (dynamicSegment is _Parameter<T>) {
+        parameters[dynamicSegment.symbol] = segment;
+        final result = next(dynamicNode, newMap);
+        if (result != null) return result;
+        parameters.remove(dynamicSegment.symbol); // backtrack
```
FXschwartz: Looks like this drops the outer binding when the same param name is bound at two levels and the inner branch fails:
```dart
final trie = PathTrie<int>()
  ..add(NormalizedPath('/:x/lit/:x/a'), 1)
  ..add(NormalizedPath('/:x/:y/c'), 2);
trie.lookup(NormalizedPath('/v/lit/c')); // params = {y: lit} — x is gone
```
On main this returns {x: v, y: lit} since the per-level copy kept the outer value. Probably a good place to add a test as well.

packages/relic_core/lib/src/router/path_trie.dart
```diff
@@ -376,23 +394,35 @@ final class PathTrie<T extends Object> {
           : (final v) => parentMap(childMap(v));
     }
     currentNode.children.addAll(node.children);
+
+    _needsBacktracking =
```
FXschwartz: I ran into what looks like a stale-flag case with `group()`. Since `attach(consume: false)` shares the subtree, routes added to the sub router afterwards mutate the parent's nodes but only seem to set the sub trie's flag:
```dart
final sub = router.group('/api');
sub.get('/users/x', 1);
sub.get('/users/:id/y', 2);
router.lookup(Method.get, '/api/users/x/y'); // PathMiss — matches {id: x} on main
```
`injectAt` looks fine since it consumes. Maybe the flag needs to live with the shared structure rather than the `PathTrie` wrapper? Not sure what the cleanest fix is. Could use a regression test too, the backtrack tests cover the happy paths but not this one.

packages/relic_core/lib/src/router/no_cache.dart
```diff
@@ -1,14 +1,7 @@
...
-/// Useful for high-cardinality workloads where caching causes more overhead
-/// than it saves (e.g., many unique dynamic paths like `/users/:id`).
-///
-/// Example:
-/// ```dart
-/// NormalizedPath.interned = NoCache();
-/// ```
+/// A no-op [Cache] that never stores or retrieves values, for opting out of
+/// caching in high-cardinality workloads where it costs more than it saves.
 final class NoCache<K, V> implements Cache<K, V> {
```
FXschwartz, line 5: Is `NoCache` still needed? I couldn't find any consumers now that `NormalizedPath.interned` is gone, so there doesn't seem to be anything to plug it into.

packages/relic_core/test/router/no_cache_test.dart
```diff
@@ -28,68 +28,32 @@ void main() {
     });
   });
 
-  group('Given NormalizedPath with NoCache', () {
-    late Cache<String, NormalizedPath> originalCache;
+  group('Given a Router (no result cache)', () {
```
FXschwartz: nit: this group tests `Router.lookupUri` rather than `NoCache`, maybe better off in a router test file?

FXschwartz (changes requested):

packages/relic_core/lib/src/router/path_trie.dart
nielsenko: Good catch. Fix is simple enough. I'll add test as well.

packages/relic_core/lib/src/router/path_trie.dart
nielsenko: Even better catch.. This one is a bit more tricky, since we allow changes to a sub-router after `attach` (which is basically what `group` does). I think I'll rollback the `_needsBacktracking` optimization for now, since it is not sound in this case.

packages/relic_core/lib/src/router/no_cache.dart
nielsenko: I would prefer to leave it in. We still use `Cache<K,V>` for static file handling and have a `LruCache` implementation of this. Having a no-op version hanging around might come in handy. That said rolling a new one is also trivial. I'll leave as is unless you disagree strongly.

packages/relic_core/lib/src/router/path_trie.dart
nielsenko: Fixed in 001fe675

packages/relic_core/lib/src/router/path_trie.dart
nielsenko: I have dropped "_feat: Infer backtracking from the route table_" commit for now (previously c4cba4d).

packages/relic_core/test/router/no_cache_test.dart
nielsenko: This is an accident of history, but okay I'll fix.

nielsenko: Fixed in d840b0d3.

packages/relic_core/lib/src/router/path_trie.dart
nielsenko: Regression test added in deaad11

packages/relic_core/lib/src/router/path_trie.dart
nielsenko: Regression test added in deaad11

FXschwartz (approved):
LGTM! 🚀

